PR #68, opened 2021-08-22
created discount area on details page

sebastiannicolajsen:
### What
Adding the capabilities for adding discount on draft order details page

### Why
Currently users can only add discounts when creating the draft order initially, not once actually created.

### How
Reusing the code for a new draft order.

### Testing
User testing


olivermrbl:
Closed #325

PR #69, opened 2021-09-03
multiply value with 100 to convert to whole number percentage

VilfredSikker:
### What
Multiplies tax_rate with 100 to convert to whole number percentage

### Why
The database saves percentages with a value between 0-100, so to visually show it as a decimal number, we have to account for this before we call the create endpoint

### How
Multiplies tax_rate with 100 when calling the `create` endpoint


olivermrbl (changes requested):
Before review, can you change the base branch to be `develop` please :)

olivermrbl (approved):

Issue #79, opened 2021-09-06
Bug: Denominations are not included when creating a giftcard product 


pKorsholm:
**What?** 
- Variants are not included in the request to the server when creating a giftcard

olivermrbl:
Cloased #85

PR #85, opened 2021-09-07
Bug: include denominations on giftcard creation

pKorsholm:
**What?**
- Correct denomination creation for giftcards
- Routing to product page of giftcards after creation
- Adjusted denomination amount to reflect the decimal representation of prices in the database

**Why?**
- Giftcard creation crashed previously
- After creation the client navigated to `product/undefined`

**How?**
- Changed field name to use object notation for index as described in useFieldArray documentation
- Included a key for each denomination
- Created a util method for price calculation when sending data to the database for easy inclusion of additional 'no division currencies' or later changes to price calculations


pKorsholm:
solves #79

olivermrbl (approved):

Issue #86, opened 2021-09-07
Bug: Default currency for giftcards is always DKK


pKorsholm:

olivermrbl:
Closed by #87

PR #87, opened 2021-09-07
Bug/default currency for giftcards

pKorsholm:
**What?** 
- Set currency correctly for giftcard creation according to store settings
- Add display price calculation to account for no division currencies

**Why?**
- Currency was locked in to DKK when creating a giftcard 
- Display price calculation assists with uniform price calculation for the frontend


pKorsholm:
solves #86

srindom (approved):

PR #89, opened 2021-09-08
Fix/update variant modal correctly passes

sebastiannicolajsen:
### What
- Allow users to create new variants (also automatically show the creation window once user clicks ‘add variant’)

### Why
- This was currently not functioning, but intended to.

### How
- Reuse existing code which allows to add copies of variants (worked fine). 

### Testing
- User testing

### DIFF
--- a/src/components/variant-grid/index.js
+++ b/src/components/variant-grid/index.js
@@ -248,98 +248,102 @@ const VariantGrid = ({ product, variants, onChange, edit, onEdit, onCopy }) => {
           </tr>
         </TableHead>
         <tbody>
-          {variants.map((v, row) => (
-            <tr key={row}>
-              {columns.map((c, col) => (
-                <Td
-                  key={`${row}-${col}`}
-                  data-col={col}
-                  data-row={row}
-                  dragover={isDraggedOver({ col, row })}
-                  onDragEnter={handleDragEnter}
-                  onClick={() =>
-                    !c.readOnly &&
-                    setSelectedCell({
-                      field: c.field,
-                      value: v[c.field],
-                      row,
-                      col,
-                    })
-                  }
-                  selected={
-                    selectedCell.row === row && selectedCell.col === col
-                  }
-                  head={c.headCol}
-                >
-                  {!(selectedCell.row === row && selectedCell.col === col) &&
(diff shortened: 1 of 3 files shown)


src/components/variant-grid/index.js
```diff
@@ -248,98 +248,102 @@ const VariantGrid = ({ product, variants, onChange, edit, onEdit, onCopy }) => {
           </tr>
         </TableHead>
         <tbody>
-          {variants.map((v, row) => (
```
sebastiannicolajsen, line 251: Not a lot has changed in this file - think i moved something slightly causing the massive markings. The relevant part is the addition of a check if the variant has even been created. If not (i.e. ID === undefined) we want to hide it until it is completed.

src/domain/products/details/variants/index.js
```diff
@@ -62,6 +62,15 @@ const Variants = ({
...
 
+  useEffect(() => {
+    if (
+      variants &&
+      variants[0] &&
+      variants[variants.length - 1].id === undefined
+    )
+      setNewVariant(variants[variants.length - 1])
+  }, [variants])
+
```
sebastiannicolajsen, line 73: Whenever we want to create a new variant we add a blank one to the list of variants, and then hereafter we must select it for opening - which is why we do this

src/domain/products/details/variants/index.js
```diff
@@ -210,6 +219,9 @@ const Variants = ({
             else if (editVariant) handleUpdateVariant(data)
           }}
           onClick={() => {
+            if (newVariant) {
```
sebastiannicolajsen, line 222: If we for some reason cancel, remove the empty one not to flood the list

src/domain/products/details/variants/variant-editor.js
```diff
@@ -372,7 +372,7 @@ const VariantEditor = ({
             Cancel
           </Button>
           <Button type="submit" variant="deep-blue">
-            Save
+            {isCopy ? "Create" : "Save"}
```
sebastiannicolajsen, line 375: Slightly change text

olivermrbl (approved):

PR #99, opened 2021-09-14
Feat/toggle adjust inventory on receive return

sebastiannicolajsen:
### What
- Allow the user to flag whether or not a return (when received) should restock inventory

### How
- Add an option to the `receive-return` menu.


olivermrbl:
Should be refactored to fit the new revamped admin. Created an issue #377.

PR #100, opened 2021-09-14
Feat/allow product selection on discounts allocated to a specific item

sebastiannicolajsen:
### What 
- Allow admins to create (and update) discounts which only apply to certain items.

### Why
- Item specific discounts have earlier been part of the product, and is an important part of creating discounts.

### How
- Make the user able to specify the items which a given discount should be valid for by extending both the `new discount` page and `discount-rule` page using a new component `product-selection` which heavily replicates the selection menu found for selecting variants when creating a `new draft order`.


olivermrbl (changes requested):
There are some UI flaws, that needs fixing:

**These two paragraphs seem to be contradictory.**
image

**These should align on one line. Consider making the selector wider or use an `...` when title overflow.**
Screenshot 2021-10-04 at 08 55 35

src/domain/discounts/details/discount-rule/index.js
```diff
@@ -249,6 +243,26 @@ const DiscountRuleModal = ({
               <Text fontWeight="500">Item</Text>
             </Pill>
           </Flex>
+          {isAllocatedToItem && (
```
olivermrbl, line 246: You should also me allowed to choose items even though the allocation method is total. This would just mean, that if you have two items in the cart where only one is applicable for the applied discount, then only this would be reflected in the total discount. This is support on the backend.

src/domain/discounts/details/index.js
```diff
@@ -338,10 +339,46 @@ const DiscountDetails = ({ id }) => {
             </Box>
           </Box>
           <Divider m={3} />
-          <Box>
-            <Text ml={3} mb={2}>
-              Applicable for all products
-            </Text>
+
+          <Box px={3}>
+            {discount.rule.allocation !== "total" &&
```
olivermrbl, line 344: See above comment. Applicable products should be available regardless of allocation method.

src/domain/discounts/new/index.js
```diff
@@ -405,55 +397,72 @@ const NewDiscount = ({}) => {
...
-            <Flex width={[1, 1, 1, 1 / 2]} flexDirection="column">
-              <StyledLabel pb={2} style={{ fontWeight: 500 }}>
-                End date
-              </StyledLabel>
-              <DatePicker
-                date={endDate}
-                onChange={setEndDate}
-                enableTimepicker={true}
+        </Flex>
+        {isAllocatedToItem && (
```
olivermrbl, line 401: See above comment.

PR #102, opened 2021-09-16
Feat:discount code expiration

pKorsholm:
This PR is dependant on [Feat:discount expiration date](https://github.com/medusajs/medusa/pull/403) in medusa core

**What?**
- Add ui for selecting starts_at and ends_at dates during discount creation and editing
- Add ui for picking a duration in discount creation and editing if the discount is dynamic

**Why?**
- To allow admins to create limited time discounts and offers

**How?**
- Adding a new component for picking a duration
- Conditionally including timestamps in the react-datepicker component

**Testing**
Manual testing

### DIFF
--- a/src/domain/discounts/new/index.js
+++ b/src/domain/discounts/new/index.js
@@ -125,8 +125,6 @@ const NewDiscount = ({}) => {
   }
 
   const submit = async data => {
-    // const iso8601Date = `P${durationYears}Y${durationMonths}M${durationDays}DT${durationHours}H${durationMinutes}M`
-
     if (isFreeShipping) {
       const disc = constructFreeShipping(data)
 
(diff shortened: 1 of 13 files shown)


src/domain/discounts/new/index.js
```diff
@@ -131,31 +143,41 @@ const NewDiscount = ({}) => {
...
+      ends_at: endDate ? moment(endDate).format("MM/DD/YYYY HH:mm") : undefined,
       regions: data.regions || [],
+      valid_duration: isDynamic ? iso8601Date : undefined,
     }
 
     if (data.usage_limit) {
       discount.usage_limit = data.usage_limit
     }
 
+    console.log(discount)
```
srindom, line 169: Remove this :)

srindom (changes requested):
The alignment here needs to be readjusted to match the design in Figma - think it is the margin bottom on the tooltip icon that does it: 
image

UI for editing discounts needs to be changed:
image

src/domain/discounts/new/index.js
pKorsholm: yes, that seems redundant, I have removed it 😄

srindom (approved):

PR #103, opened 2021-09-16
Feat/note on order

sebastiannicolajsen:
### What

- Add the ability to add, modify and delete notes related to an order

### Why

- This may be usable for the administrators wanting to add notes related to orders

### How
- Implement a note timeline and create option and create the ability to call this from the api services


srindom:
image

Final change and we should be good -> please show the email in case the author doesn't have first and last names populated. If there is no author then "Unknown" is fine.

srindom (changes requested):
small style fix

PR #107, opened 2021-09-21
feat/customer-information

sebastiannicolajsen:
### What
- Update admin view of customer details to enhance the experience as well as giving them the option to update `first_name`, `last_name`, `phone` and `email` (`email` only being if the customer has not signed up with the account).

### Why
- This is useful for administrators especially when a user may not have created their account themselves or signed up.

### How
- Modify `customer/details` and create an `edit` component.

### DIFF
--- a/src/domain/customers/details/index.js
+++ b/src/domain/customers/details/index.js
@@ -10,6 +10,7 @@ import {
   TableRow,
   DefaultCellContent,
   BadgdeCellContent,
+  TableLinkRow,
 } from "../../../components/table"
 import styled from "@emotion/styled"
 import ReactJson from "react-json-view"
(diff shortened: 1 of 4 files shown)


src/domain/customers/details/index.js
```diff
@@ -69,27 +112,85 @@ const CustomerDetail = ({ id }) => {
...
+                        color={decideBadgeColor(order.payment_status).color}
+                        bg={decideBadgeColor(order.payment_status).bgColor}
+                      >
+                        {order.payment_status}
+                      </Badge>
+                    </BadgdeCellContent>
+                  </TableDataCell>
+                  <TableDataCell>
+                    <DefaultCellContent>
+                      {order.total} {order.currency_code.toUpperCase()}
```
olivermrbl, line 187: Use our util function for prices please.

olivermrbl (changes requested):

olivermrbl (changes requested):
The prices are still displaying wrong. Please see attached image and correct it :)
image

Also, only the order number should be blue.

olivermrbl:
> The prices are still displaying wrong. Please see attached image and correct it :) image
> 
> Also, only the order number should be blue.

The content of the rows is still blue :) Should only be the display order number as mentioned previously. @sebastiannicolajsen

olivermrbl (approved):

PR #110, opened 2021-09-28
Feat:swap/claim on claim and claim on swap

sebastiannicolajsen:
### What
- Allowing claims to be claimed and swapped, and allow swaps to be claimed. 

### Why
- Claims may be required to be swapped or claimed (e.g. wrong size, missing item)

### How
- Make sure to include correct claim items in the create swap and claim components.


src/domain/orders/details/utils/create-filtering.js
```diff
@@ -1,4 +1,4 @@
-export const filterItems = (order, isClaim) => {
+export const filterItems = order => {
```
olivermrbl, line 1: It seems, that I cannot receive the items from the claim here in the admin. Look at the video below. It only allows me to receive a single item.
https://user-images.githubusercontent.com/59018053/135453863-cfd50153-bab6-4408-be19-14c6d75cee81.mov

olivermrbl (changes requested):

src/domain/orders/details/returns/receive-menu.js
```diff
@@ -221,6 +227,8 @@ const ReceiveMenu = ({
               return
             }
 
+            console.log(item)
+
```
olivermrbl, line 231: rm

olivermrbl:
Solved #325

PR #112, opened 2021-09-28
Fix: expandable sidebar issues

sebastiannicolajsen:
### What
- Only allow one category to ever be expanded at once in the sidebar, and only collapse this when another element is expanded.

### Why
- To reduce clutter and ease of use

### How
- Implement custom trigger behaviour and apply this to all collapsables. (Cannot generate custom function component to increase reuse as this removes the animations) Also, this includes wrapping elements without subitems in a collapsible.

### DIFF
--- a/src/components/sidebar/index.js
+++ b/src/components/sidebar/index.js
@@ -89,6 +89,8 @@ const Sidebar = ({}) => {
       handleTriggerClick: () => setCurrentlyOpen(id),
     }
   }
+  // We store the `id` counter on the function object, as a state creates
+  // infinite updates, and we do not want the variable to be free floating.
   triggerHandler.id = 0
 
   return (
--- a/src/components/sidebar/index.js
+++ b/src/components/sidebar/index.js
@@ -63,6 +63,7 @@ const Sidebar = ({}) => {
   const [storeName, setStoreName] = useState("")
   const [path, setPath] = useState("")
   const [orderOpen, setOrderOpen] = useState(false)
+  const [currentlyOpen, setCurrentlyOpen] = useState("")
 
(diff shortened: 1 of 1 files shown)


src/components/sidebar/index.js
```diff
@@ -81,6 +82,15 @@ const Sidebar = ({}) => {
...
   }, [])
 
+  const triggerHandler = () => {
+    const id = triggerHandler.id++
+    return {
+      open: currentlyOpen === id,
+      handleTriggerClick: () => setCurrentlyOpen(id),
+    }
+  }
+  triggerHandler.id = 0
```
olivermrbl, line 92: What does this line do? Would be great to move this to a place, that makes a bit more sense or add a comment :)

olivermrbl: @sebastiannicolajsen

olivermrbl, line 92: This requires a comment or two. What happens and is this necessary?

olivermrbl (changes requested):

src/components/sidebar/index.js
olivermrbl: @sebastiannicolajsen

sebastiannicolajsen: Should I add the comment in the source code? This is a generalized way of generating the two props `open` and `handleTriggerClick` Which allows us to implement custom behaviour on clicking the sidebar elements. Specifically, we want to only have one collapsible open at a time, which is why they all need a unique identifier which is selected as the one open (id). To ensure no conflict with other variables, we store the id on the function object.

olivermrbl (approved):

PR #115, opened 2021-09-29
fix: dropdowns on top

sebastiannicolajsen:
### What
- Changing dropdowns (`dropdown` and `react-select` components) are on top of modals (and other things). 

### Why
- These sometimes overflow the modal resulting in the user needing to scroll to select input. This is a nuisance for the user.

### How
- Wrap the `dropdown`'s `scrollable` so that itss absolute position is respected. Ensure that `menuPortalTarget` is always set to `document.body` for the `react-select` and that this is closed on resizing as to avoid random floating behaviour of the `scrollable`.

### DIFF
--- a/src/components/dropdown/index.js
+++ b/src/components/dropdown/index.js
@@ -14,7 +14,6 @@ export const DropdownContainer = styled(Box)`
 
   transform: ${props =>
     props.dropUp ? "translateY(-180px)" : "translateY(32px)"};
-  position: absolute;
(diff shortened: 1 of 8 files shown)


olivermrbl (approved):
Very nifty!

olivermrbl:
@sebastiannicolajsen Reverted this commit. We were experiencing issues with the order details overview. The dropdowns pretty much floated all over the page :)

PR #117, opened 2021-09-30
Feat: admin error handling

sebastiannicolajsen:
### What
- Display the error messages given by the API in toasters across the admin project.

### Why
- The current error messages were providing less feedback than these, and utilising the errors given by the API provides greater flexibility in feedback.

### How
- Add a utility function to extract error messages and use this across the project to fetch error messages for toasters. Doing so, allows us to unify (most) error messages / display strings.


src/domain/gift-cards/detail.js
```diff
@@ -68,10 +69,8 @@ const GiftCardDetail = ({ id }) => {
   const { regions } = useMedusa("regions")
 
   const formatNumber = n => {
```
pKorsholm, line 71: does this not exist in utils/prices

PR #119, opened 2021-09-30
Feat: product filtering

pKorsholm:
**What**
- Product filtering similar to that of orders 

**How**
- using the same patterns and components as was used in orders

### DIFF
--- a/src/domain/products/index.js
+++ b/src/domain/products/index.js
@@ -5,6 +5,7 @@ import { Flex, Text, Box, Image } from "rebass"
 import { Input } from "@rebass/forms"
 import { Router } from "@reach/router"
 import Medusa from "../../services/api"
+import ProductsFilter from "./filter-dropdown"
 
 import ImagePlaceholder from "../../assets/svg/image-placeholder.svg"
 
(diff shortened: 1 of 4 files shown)


src/domain/products/index.js
```diff
@@ -246,6 +342,20 @@ const ProductIndex = () => {
...
+          setTagsFilter={setTagsFilter}
+          submitFilters={submit}
+          tagsFilter={tagsFilter}
+          filterTags={filterTags}
+          setFilterTags={setFilterTags}
+          resetFilters={resetFilters}
+          clearFilters={clearFilters}
+        />
         {selectedProduct && (
           <Button
```
olivermrbl, line 360: Add `ml={2}` to the "New product" button :)

src/domain/products/index.js
```diff
@@ -68,17 +78,98 @@ const ProductIndex = () => {
...
+        searchObject.limit = 20
+      }
+
+      const query = qs.stringify(queryObject)
+      window.history.replaceState(`/a/products`, "", `${`?${query}`}`)
+      refresh({ search: { ...searchObject } })
+    }
+  }
+
+  const clearFilters = () => {
```
olivermrbl, line 168: Upon clearing filters, we should use the default offset and limit. Currently, we fetch all products.

olivermrbl (changes requested):
A couple of changes to the UI and one fix to how we fetch tag-usage.

We are currently fetching `tag-usage` quite a lot. Both on initial load and everytime we click a button. We should limit this to a single call upon opening the products overview. 

See attached image for reference: 
image

olivermrbl:
> A couple of changes to the UI and one fix to how we fetch tag-usage.
> 
> We are currently fetching `tag-usage` quite a lot. Both on initial load and everytime we click a button. We should limit this to a single call upon opening the products overview.
> 
> See attached image for reference: image

Did you get a chance to look at these as well? @pKorsholm

PR #123, opened 2021-10-05
Fix/modal on top

sebastiannicolajsen:
### What
- Changing dropdowns (`dropdown` and `react-select` components) to be on top of modals (and other things). This is always true for `react-select` but is required to be toggled on for `dropdown`.

### Why
- These sometimes overflow the modal resulting in the user needing to scroll to select input. This is a nuisance for the user.

### How
- Wrap the `dropdown`'s `scrollable` so that its absolute position is respected (when float is set to `true`). Ensure that `menuPortalTarget` is always set to `document.body` for the `react-select` and that this is closed on resizing as to avoid random floating behaviour of the `scrollable`.

### DIFF
--- a/src/components/dropdown/index.js
+++ b/src/components/dropdown/index.js
@@ -14,7 +14,6 @@ export const DropdownContainer = styled(Box)`
 
   transform: ${props =>
     props.dropUp ? "translateY(-180px)" : "translateY(32px)"};
-  position: absolute;
   background-color: #fefefe;
   min-width: ${props => props.dropdownWidth};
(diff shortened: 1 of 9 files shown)


olivermrbl:
We've decided to rethink the approach to dropdowns generally since the current implementation was bugging us in different edge cases, that could not be solved.

Briefly, the solution uses a React.Portal to arrach the dropdown element next to the root HTML container. This gives us full control of the positioning of the dropdown and allows us to keep scrollable parents. A PR will be opened soon.

PR #126, opened 2021-10-06
Feat: Unified search bar

sebastiannicolajsen:
### What
- Implement a fixed top bar that allows the user to search within the domain of their current location (i.e. orders in the orders page, products in the products page, customers in the customer page, discounts in the discount page, and gift cards in the gift card page). 
- Allow the users to also log out using this bar.

### Why
- It unifies the look, and the option to log out what currently missing.

### How
- Implement an interface context that allows each domain to submit their own search handlers to the top bar.
- Utilise the auth session, and update the logout part of the API service to call the newly added `delete session` found in [this](https://github.com/medusajs/medusa/pull/459) PR to Medusajs


olivermrbl (changes requested):
Great work.. A couple of things. It does not seem to apply our "tab filters" with the search query. 
It work seamlessly, when searching -> adding a filter from the dropdown. 
But does not work, when search -> adding a tab filter

Additonally, the search bar should be disabled or removed from the settings page.

Issue #128, opened 2021-10-07
Fulfillment Provider mapper should return a default object when id is unknown


srindom:
If the fulfillment provider id is not known by [the fulfillment mapper](https://github.com/medusajs/admin/blob/master/src/utils/fulfillment-providers.mapper.js), it returns an empty string. It should return a default value that just has:
```
{
  label: provider,
  value: provider
}
```

Like in [the payment provider mapper](https://github.com/medusajs/admin/blob/master/src/utils/payment-providers-mapper.js).

olivermrbl:
Closed #129

PR #129, opened 2021-10-07
fix: Update default return value for fulfillment mapper

bsantosh909:
### What

- If the fulfillment provider id is not known by the fulfillment mapper, it returns an empty string

### Why

- This is done as requested in https://github.com/medusajs/admin/issues/128

### How

- Just like pointed in the original Issue the provider value is returned for default case
 ```js
{
  label: provider,
  value: provider
}
```


olivermrbl (approved):
Great work!

PR #130, opened 2021-10-08
Feat: new login design

sebastiannicolajsen:
### What
- Update login page to accommodate the new design.

### Why
- We should utilise the new design across the platform.

### How
- Reuse components from Medusa cloud (and adapt these).


olivermrbl (changes requested):
See comment. Figma file is the top image of the two. 

Additionally, did you branch out from develop, cause I see a couple of things that are not related to new login? :)

olivermrbl:
Included in #131

PR #136, opened 2021-10-11
feat: add custom shipping options

zakariaelas:
### What 

- Admin support for adding custom shipping options to swaps when creating them

### Why 

- Allow store managers to set custom shipping options on swaps

### How

- Adds color and text themes from medusa-cloud
- Tweaks `components/button/index.js` to handle case where button is `deep-blue` 
- Tweaks design to follow Figma mockups
- Implements functionality for setting custom shipping options and sending them as part of the payload to `POST /orders/:id/swaps`


src/domain/orders/details/swap/create.js
```diff
@@ -73,6 +80,36 @@ const extractPrice = (prices, order) => {
...
+    0
+  )
+  return `${total.toFixed(2)} ${order.currency_code.toUpperCase()}`
+}
+
+const prepareCustomShippingOptions = (
+  customShippingOptions,
+  shippingOptions
+) => {
+  if (!customShippingOptions.length) return []
```
pKorsholm, line 95: Flavor thing(but it's what i can comment on 😉 ), I think this should be in curly braces

src/domain/orders/details/swap/create.js
```diff
@@ -532,4 +763,54 @@ const SwapMenu = ({ order, onCreate, onDismiss, toaster }) => {
   )
 }
 
+const CurrencyInput = ({ value, currency, ...props }) => {
```
pKorsholm, line 766: I dont know if it's convention but would it make sense to group these with the rest of the styled components at the top of the file

src/domain/orders/details/swap/swap-details/shipping-methods.js
```diff
@@ -21,7 +21,7 @@ const SwapShippingMethods = ({ shipping_methods = [], taxRate, currency }) => {
                   <>
                     {method.shipping_option.name} -{" "}
                     {formatDecimalAndCurrency(
-                      ((100 + taxRate) * method.shipping_option.amount) / 10000,
```
pKorsholm, line 24: There is a util method in utils/prices.js for calculating this and checking for noDivisionCurrencies

PR #140, opened 2021-10-13
added quantity to calculate total price

sebastiannicolajsen:
### What
- Ensure that all `requirements` on a `shipping_option` respected when listing options for a given draft order.

### How
- Turns out we were missing to include quantity in our calculation of total price in the new draft order component.

### Testing
- Tested that requirements were adhered to when: (1) one `min_subtotal` requirement was present, (2) one `max_subtotal` requirement was present, (3) one of each were present, and (4) none were present.

### DIFF
--- a/src/domain/orders/new/new-order.js
+++ b/src/domain/orders/new/new-order.js
@@ -319,9 +319,9 @@ const NewOrder = ({ onDismiss, refresh }) => {
   const calculateTotal = () => {
     const tot = items.reduce((acc, next) => {
       if ("unit_price" in next) {
-        acc = acc + next.unit_price
+        acc = acc + next.unit_price * next.quantity
       } else {
-        acc = acc + extractUnitPrice(next, region, false)
+        acc = acc + extractUnitPrice(next, region, false) * next.quantity
       }
 
       return acc


olivermrbl (changes requested):
_edit: Great!_

olivermrbl:
> I am quite sure, that `fetchShippingOptions` needs some changes as well. Especially the check for requirements. Something along the lines of:
> 
> ```
> async fetchShippingOptions() {
> ...
> // Filter out options by requirements
>       const options = data.shipping_options.reduce((acc, next) => {
>         if (next.requirements?.length) {
>           const minSubtotal = next.requirements.find(
>             req => req.type === "min_subtotal"
>           )
> 
>           if (minSubtotal) {
>             if (total >= minSubtotal.amount) acc.push(next)
>           }
> 
>           const maxSubtotal = next.requirements.find(
>             req => req.type === "max_subtotal"
>           )
>           if (maxSubtotal) {
>             if (total <= maxSubtotal.amount) acc.push(next)
>           }
>         } else {
>           acc.push(next)
>         }
> 
>         return acc
>       }, [])
> ...
> }
> ```

And actually also need a check for scenarios in which the option has both types of requirements

PR #143, opened 2021-10-16
Create formatNumber utility helper function

### LINKED ISSUE #141: Gift-cards detail component does not respect no division currencies
Specifically, the `formatNumber` does the recalculation itself instead of using separate utility functions.

askalik:
Resolves #141 takes formatNumber and moves to utility function / imports from utils


olivermrbl:
Thanks for the contribution. Solved in #325 💜

PR #161, opened 2021-10-18
fix: dropdown behavior scroll containers

zakariaelas:
### What 

- Fixes dropdowns' overflow behavior across the codebase

### Why

- Better UI/UX

### How
- Use react-popper, which is a react adaptation of popper.js, a positioning engine for tooltips/popovers. Instead of having to handle various corner cases related to calculating the position of the dropdown. It's better to leverage this library as it makes our solution more scalable and reliable
- Updated occurences of the `Dropdown` component across the codebase to make sure the right props are passed


olivermrbl (approved):
Great work!

PR #175, opened 2021-11-22
fix: add missing lodash dependency

jaredhill4:
While working on the Gatsby theme we are creating for the admin, we noticed that `lodash` was not included in the package dependencies, and it was not imported in `src/domain/products/new/index.js`.

In this PR, I've added `lodash` to the dependencies and added the missing import.


olivermrbl (approved):
Thanks for the fix! 👍

Issue #178, opened 2021-11-28
wrong css selector syntax in StyledMultiSelect


Manoj0Marmat:
**Issue Type** : Bug

**Error Occurs in** : 8 files in index.js

**Reason of Error** : Wrong syntax of css selector

Manoj0Marmat:
@olivermrbl fixed

olivermrbl:
Closed #325

PR #179, opened 2021-11-28
fix: css syntax issue

Manoj0Marmat:
fix syntax issue #178  by removing unwanted extra colon after css selector


olivermrbl:
Thank you for the contribution. The issue is resolved as part of #325

PR #182, opened 2021-11-30
update node packages (gatsby, react, axios) and fix dropdown ref

jaredhill4:


src/components/dropdown/index.js
```diff
@@ -129,7 +129,6 @@ const Dropdown = ({
         dropUp={dropUp}
         leftAlign={leftAlign}
         minWidth={dropdownWidth || "160px"}
-        ref={ref}
```
jaredhill4, line 132: For some reason, after updating React, the placement of this `ref` was no longer working. (To be honest, I'm not sure how it ever was working, since the trigger `Button` is outside of it and the `handleClickOutside` function should have been running every time you clicked the button, thus causing the dropdown to immediately hide.)

jaruesink:
@jaredhill4 they might not want to switch to yarn, should probably remove the package lock diffs

jaredhill4:
> @jaredhill4 they might not want to switch to yarn, should probably remove the package lock diffs

@jaruesink - I actually REMOVED the `yarn.lock` file as part of this PR. (I had added it in my last PR).

Issue #176, opened 2021-11-23
Auto-fill customer details when creating draft orders from admin


srindom:
When we have the customer data already we should auto-fill the address form in the "Create Draft Order" modal

Manoj0Marmat:
@srindom assign this to me

srindom:
@Manoj0Marmat how is it going with this? :)

Manoj0Marmat:
@srindom understanding code

Manoj0Marmat:
@srindom done

Manoj0Marmat:
@srindom dine check

Manoj0Marmat:
@srindom will update for address, province , post code and city details

Manoj0Marmat:
@srindom 

https://user-images.githubusercontent.com/85970385/144227346-8943669f-9ae9-4396-bd22-3a3e95f02aa7.mov

olivermrbl:
Solved in #325

PR #185, opened 2021-12-01
feat : Auto-fill customer details when creating draft orders from admin

Manoj0Marmat:
#176 auto-fill first name, last name and phone details in the "Create Draft Order" modal

### DIFF
--- a/src/domain/orders/new/components/shipping-details.js
+++ b/src/domain/orders/new/components/shipping-details.js
@@ -21,7 +21,7 @@ const AddressContainer = styled(Flex)`
   `}
 `
 
-const ShippingDetails = ({ customerAddresses, setCustomerAddresses, form }) => {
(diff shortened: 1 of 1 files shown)


srindom:
Great work so far - can you fill out address details too? 

Customers have a `shipping_addresses` field - I would suggest that you filter the addresses based on whether the `draft_order.region.countries` include the `shipping_address.country_code` and then if the resulting array only has 1 address we fill that out; if there are more we don't do anything (for now; thinking we can add a nice selector UI later).

Manoj0Marmat:
@srindom 
Screenshot 2021-12-01 at 2 57 07 PM
 
this is details of selected customer can we get shipping_addresses stuffs using costumer id ?

Manoj0Marmat:
@srindom hello sir i'm trying to contact you on discord username Li_0(manoj)

Issue #198, opened 2021-12-16
Reference error: navigate is not defined (order details)


Orick08:
When you open an order you can see the list of the items that order have:
<img width="238" alt="Screen Shot 2021-12-16 at 1 05 09 PM" src="https://user-images.githubusercontent.com/33299343/146441605-21f44ad7-7471-4db4-a25d-4921085a2f3a.png">

When you click it should redirect to the item but instead it triggers an error:
<img width="1039" alt="Screen Shot 2021-12-16 at 12 34 44 PM" src="https://user-images.githubusercontent.com/33299343/146441676-dfc19223-22f5-4556-8303-27207266c2d7.png">

Orick08:
This issue is only on master branch, is already solved in develop

PR #199, opened 2021-12-16
Fix: missing import navigate from gatsby

Orick08:
What - Adds an import on src/domain/orders/details/timeline/index.js

Why - This prevents reference error while you click a product of the timeline.

Testing - The link no longer triggers the error and redirects to the correct page `/a/products/${productId}`

This solve the issue #198


Orick08:
Ups, this should commit into development, not master

Issue #72, opened 2021-09-06
Bug: General validation in forms


pKorsholm:
**What?**
- Forms need general validation of required fields before submission

rolwin100:
@pKorsholm can u list all the places where the validations are needed.

pKorsholm:
The idea is to highlight which fields are missing to the user so they can easily identify which fields are missing data

Ideally this would apply to any form in the admin portal with required fields, I might have missed some, but here is a list:
- When creating an order
- When creating a discount
- Product creation/update
- Region details

Thanks a lot Rolwin 😄

rolwin100:
ur welcome @pKorsholm :)

PR #200, opened 2021-12-17
Fix: invalid login feedback

Orick08:
## What?
Shows a message to the user if the email or password is wrong. This solves the issue #181.
## Why?
In the current version, if the credentials are wrong it only shows the login again without any feedback.
## How?
I add a boolean flag on the page, but for further forms validations (issue #72) maybe is better to discuss a library to do this automatically like formik-yup.
## Testing?
I tested with invalid credentials, the only misbehave is that if the server is down and cant communicate it shows the same message.
## Screenshots
Screen Shot 2021-12-17 at 12 08 51 PM

### DIFF
--- a/src/pages/login.js
+++ b/src/pages/login.js
@@ -28,11 +28,11 @@ const IndexPage = () => {
       .then(() => {
         navigate("/a")
       })
-      .finally(() => {
-        setLoading(false)
-      })
       .catch(e => {
         setIsInvalidLogin(true)
+      })
+      .finally(() => {
+        setLoading(false)
       })
   }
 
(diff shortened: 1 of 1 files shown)


olivermrbl (approved):
LGTM - thanks for your contribution! ❤️

src/pages/login.js
```diff
@@ -29,6 +31,9 @@ const IndexPage = () => {
       .finally(() => {
         setLoading(false)
       })
+      .catch(e => {
+        setIsInvalidLogin(true)
+      })
```
olivermrbl, line 36: Tiny thing. I think we should change to have the `finally`-block as the last part of the promise chain.

olivermrbl (changes requested):

src/pages/login.js
Orick08: Yep, no problem, done

olivermrbl:
Thanks for the fix @Orick08 ! We included a very similar fix in #325 so closing this one :)

Issue #154, opened 2021-10-18
Missing save or cancel button on add note in timeline


sebastiannicolajsen:
While it is possible to create a note by pressing enter, it is currently not possible to cancel the note as neither a save or cancel button as is proposed on the figma design:

![Skærmbillede 2021-10-18 kl  15 40 33](https://user-images.githubusercontent.com/80953876/137742958-db0c2b35-47dd-4b88-abce-0dc7145136be.png)

olivermrbl:
Closed #201

PR #201, opened 2021-12-17
Fix: Add cancel save in timeline notes

Orick08:
## What?
Show controls for cancel/save notes in the timeline of the orders. This solves the issue #154.
## Why?
In the current version, is impossible to cancel the notes and the options are blocked until you write a note.
## How?
The two buttons were added, the function of `handleCancelNote` was already in the code but never implemented.
## Testing?
I tested canceling the note, adding the note with the save button, and pressing enter.
## Screenshots
Screen Shot 2021-12-17 at 4 45 08 PM

### DIFF
--- a/src/domain/orders/details/index.js
+++ b/src/domain/orders/details/index.js
@@ -521,7 +521,7 @@ const OrderDetails = ({ id }) => {
       })
   }
 
-  const handleCreateNote = event => {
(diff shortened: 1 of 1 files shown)


olivermrbl (approved):
LGTM, thanks! 👍

PR #208, opened 2021-12-29
feat: revamp buttons

kasperkristensen:
As I was implementing some of the new buttons design for the two split body pane, I thought it would make better sense to separate them into a PR of their own and implement all the styles + a `<Button />` component.

**What**

- Adds the three button styles as reusable Tailwind  component styles in `globals.css`.
- Adds a new atom Button component


olivermrbl (commented):
Think this is a great idea. Let's go ahead and delete those CSS files right away :)

src/components/atoms/button.js
```diff
@@ -0,0 +1,45 @@
...
+    return (
+      <button
+        className={`btn-${variant}-${size} ${className}`}
+        onClick={handleClick}
+        disabled={disabled || loading}
+        type={type}
+        ref={ref}
+        {...props}
+      >
+        {Children.map(children || null, (child, i) => {
```
olivermrbl, line 33: nit: I am quite sure, that `Children.map` will handle `children` being `null` gracefully

src/components/atoms/button.js
```diff
@@ -0,0 +1,45 @@
+import React, { Children } from "react"
+
+const Button = React.forwardRef(
```
olivermrbl, line 3: Would argue that we might as well convert the button to TS now. What do you think?

olivermrbl (commented):
Great! Couple of comments :) @kasperkristensen

olivermrbl:
And please rebase off of `feat/revamp` :)

src/components/atoms/button.js
kasperkristensen: Definitely! :)

src/components/fundamentals/button/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+        break
+      default:
+        style = "btn-primary-large"
+        break
+    }
+
+    return (
+      <button
+        {...attributes}
+        className={classNames(style, attributes.className)}
```
zakariaelas, line 66: tiny nit: maybe a good candidate for using [clsx](https://www.npmjs.com/package/clsx). Current approach is totally valid so feel free to ignore! Just a small heads-up as we'll probably be constructing classes conditionally in the future

zakariaelas (approved):
LGTM w. very tiny nit

src/components/fundamentals/button/index.tsx
olivermrbl: Are there any significant differences between [`classNames`] and `clsx` @zakariaelas? As far as I can see, they both support conditional classes in an almost identical manner.

zakariaelas: Correct! As things stand right now, the two would behave the exact same way. Mostly a matter of preference and highly opinionated as personally, i'm not the biggest fan of ternary operators inside `className`.

Only difference could possible be down the line. Consider this example: 

```jsx
// without clsx
const SomeComponent = () => {
  // some state .. using `isOpen` here but could be pretty much anything
  const [isOpen, setIsOpen] = useState(false)
  return (
    <div className={`class-1 class-2 ${isOpen ? "class-open" : ""}`}>
      ...
    </div>
  )
}

// with clsx
const SomeComponent = () => {
  const [isOpen, setIsOpen] = useState(false)
  return (
    <div className={clsx("class-1", "class-2", { "class-open" : isOpen })}>
      ...
    </div>
  )
}
```
No significant difference here ... Just a tiny utility library that would help compose `className` conditionally, especially when having class names depending on multiple pieces of state. Really, feel free to ignore and merge :sweat_smile:

olivermrbl: I am all in for it! Think it's very nifty and will be very handy in the future 😄

src/components/atoms/spinner.tsx
```diff
@@ -0,0 +1,35 @@
...
+  size = "large",
+  variant = "primary",
+}) => {
+  return (
+    <div
+      className={clsx(
+        "flex items-center justify-center",
+        { "h-[24px] w-[24px]": size === "large" },
+        { "h-[20px] w-[20px]": size === "medium" },
+        { "h-[16px] w-[16px]": size === "small" }
```
olivermrbl, line 19: 😍

src/components/organisms/topbar.tsx
```diff
@@ -1,12 +1,16 @@
...
 import NotificationBell from "../molecules/notification-bell"
 import SearchBar from "../molecules/search-bar"
 
 const Topbar: React.FC = () => {
   return (
     <div className="w-full min-h-topbar max-h-topbar pr-xlarge pl-base bg-grey-0 border-b border-grey-20 sticky top-0 flex items-center justify-between z-50">
       <SearchBar />
+      <Button variant="primary" size="large">
+        Heeey
+      </Button>
```
olivermrbl, line 13: I guess it's for testing purposes, but should we omit it from the PR 😅

olivermrbl (commented):
LGTM

src/components/fundamentals/button/index.tsx
kasperkristensen: Seems super nifty I've replaced the utility method with `clsx` instead. Also removed the switch statement, as the result could be achieved by simply doing `"btn-" + variant + "-" + "size"`. I Tailwind just does not like template literals, since doing `{`btn-${variant}-${size}`}` results in the style not being loaded 😅

src/components/organisms/topbar.tsx
kasperkristensen: 😅 fixed

zakariaelas: come on oli .. it's called being polite to users :laughing:

olivermrbl: Haha 😆

Issue #195, opened 2021-12-13
Hide swap payment in case of cancelled swaps


olivermrbl:
In the payment overview on order pages, we currently show payments for swap even though the swap has been canceled.

We need to exclude canceled swaps when doing our [summation here](https://github.com/medusajs/admin/blob/c73d11f149cc5af2af20250435a7924fadcfd1bb/src/domain/orders/details/index.js#L293).

olivermrbl:
Closed #325

PR #210, opened 2021-12-30
Hide canceled swaps from payment sum

Zeusjonass:
Hi, this is my solution for the issue #195 

In the past, when we consulted the details of an order
Screen Shot 2021-12-30 at 9 30 48

When we checked the payment overview
Screen Shot 2021-12-30 at 9 31 54

In the sum, those swaps that were canceled were considered. This should not be like that. My solution was to add a filter to exclude those swaps canceled when we make the map for the sum.

Please let me know if there is anything I can improve :)

### DIFF
--- a/src/domain/orders/details/index.js
+++ b/src/domain/orders/details/index.js
@@ -293,7 +293,9 @@ const PaymentDetails = ({ order }) => {
   }
 
   if (order?.swaps?.length) {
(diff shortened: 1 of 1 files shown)


olivermrbl (approved):
This is great! Many thanks for the contribution :)

Issue #211, opened 2021-12-30
Swap value is undefined


NayeBeckham:
**Describe the bug**
The application breaks when I'm trying to go to the swaps page because it says:
<img width="1438" alt="Screen Shot 2021-12-30 at 11 29 20" src="https://user-images.githubusercontent.com/78670199/147779677-36846629-39e5-459f-aaf3-4504f390d17b.png">

**What page/module is the bug on?**
Orders/Swaps

**To Reproduce**
Go to Orders
Click on Swaps

I was trying to contribute to the application and I was navigating on it and I saw this issue. Hope it can be useful, I'm going to do a PR to solve it.

olivermrbl:
Closed #212

PR #212, opened 2021-12-30
Fix/swaps undefined and border needs to be boolean

NayeBeckham:
Solves [#211 ](https://github.com/medusajs/admin/issues/211)

**What's the problem**
The application was breaking at the moment to go to the swaps page because the `swap` variable was undefined:
Screen Shot 2021-12-30 at 11 29 20

Also, we were having a console error in the `border` variable in the components/popover. It wasn't breaking the application but it could be fixed too.

Screen Shot 2021-12-30 at 11 46 09

**Why we should fix it**
We should fix it because the application should work as normal even if there's no data. We need to add a validation if there's no data because at the moment of use `!swaps.length if swaps is undefined it is going to break.

**How I fixed it**
Just changed the validation from `!swaps.length ?` to `swaps === undefined ?` in the swaps/index.js
And just changed `border="true"` to `border={true}` in the components/layout/popover.js
And it's working fine:
Screen Shot 2021-12-30 at 13 03 47

**Testing**
No needed.

Hope it is useful, I was navigating through the application and saw that. Tell me anything, thanks!

### DIFF
--- a/src/domain/orders/swaps/index.js
+++ b/src/domain/orders/swaps/index.js
@@ -135,7 +135,7 @@ const SwapIndex = ({}) => {
             <Spinner dark />
           </Box>
         </Flex>
-      ) : swaps === undefined ? (
+      ) : !swaps?.length ? (
         <Flex alignItems="center" justifyContent="center" mt="10%">
           <Text height="75px" fontSize="16px">
             No swaps found
--- a/src/domain/orders/swaps/index.js
+++ b/src/domain/orders/swaps/index.js
@@ -135,7 +135,7 @@ const SwapIndex = ({}) => {
             <Spinner dark />
           </Box>
         </Flex>
-      ) : !swaps.length ? (
+      ) : swaps === undefined ? (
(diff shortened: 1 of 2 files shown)


src/domain/orders/swaps/index.js
```diff
@@ -135,7 +135,7 @@ const SwapIndex = ({}) => {
             <Spinner dark />
           </Box>
         </Flex>
-      ) : !swaps.length ? (
+      ) : swaps === undefined ? (
```
olivermrbl, line 138: To follow the convention used in other domains, can we do `!swaps?.length` instead? :)

olivermrbl (commented):
Many thanks for the contribution @NayeBeckham . Tiny comment. Let me know what you think :)

src/domain/orders/swaps/index.js
NayeBeckham: Of course, thank you!

NayeBeckham: It is ready:)

olivermrbl (approved):
Great!

PR #214, opened 2022-01-03
Feat/revamp convert icons to ts

pKorsholm:
**What** 
- convert all existing icons to typescript
- include a generalised icon interface `IIconProps` with the following properties
  - `size?: string | number`
  -  `color?: string`


src/components/fundamentals/icons/search-icon.tsx
```diff
@@ -1,16 +1,17 @@
...
+const SearchIcon: React.FC<IIconProps> = ({
+  size = "24px",
+  color = "currentColor",
+}) => {
   return (
     <svg
       width={size}
       height={size}
-      viewBox="0 0 24 24"
+      viewBox={`0 0 ${size} ${size}`}
```
kasperkristensen, line 12: Small change, don't change the viewbox of Icons, as it will result in the icons becoming cropped. Changing size alone will scale the icons properly. 

Also applies to all other icons

kasperkristensen (changes requested):
nit: Don't think we should use the I<interface name> naming convention 😄

src/components/fundamentals/icons/icon-interface.tsx
```diff
@@ -0,0 +1,4 @@
+export default interface IconProps {
+  size?: string | number
+  color?: string
+}
```
kasperkristensen, line 4: One more thing, I think we should add another field to IconProps:

`attributes?: React.SVGAttributes<SVGElement>`

and then spread attributes over the `<svg>` tag, as `{...rest}` was done previously.

This way we can also pass other standard React props to the icon such as aria labels, className, etc.

kasperkristensen (changes requested):

src/components/fundamentals/icons/icon-interface.tsx
pKorsholm: sure 😄

kasperkristensen (approved):
LGTM 😄

olivermrbl (commented):
Everything looks great! What do you think about creating a `/types` similar to how we do in core to hold shared types? :) @pKorsholm

PR #215, opened 2022-01-03
fix: Remove use of old layout

olivermrbl:
Removes all files related to the old page layout. 

While migrating to Gatsby V4, some old layout functionality was tampering with the build. More specifically the profile popover. Since we don't use these files any longer, I thought I'd go ahead and remove them.


kasperkristensen (changes requested):
We should also remove `components/login-layout/layout.css` 😄

kasperkristensen (approved):
LGTM 😄

PR #216, opened 2022-01-03
feat: Migrates to Gatsby V4

olivermrbl:
Upgrades the project to Gatsby V4.

Adds webpack config to include browserified versions of libraries `stream` and `crypto` that are used by `jsonwebtoken`. This is necessary because  webpack > 5 has removed polyfills that previously handled the dirty work for us.

Should be merged after #215


package.json
```diff
@@ -46,24 +43,20 @@
...
+    "stream-browserify": "^3.0.0"
   },
   "devDependencies": {
     "@babel/core": "^7.10.2",
     "@storybook/addon-actions": "^5.3.19",
-    "@storybook/addon-backgrounds": "^5.3.19",
-    "@storybook/addon-links": "^5.3.19",
-    "@storybook/addons": "^5.3.19",
     "@storybook/react": "^5.3.19",
-    "autoprefixer": "^10.4.0",
```
kasperkristensen, line 66: PostCSS requires Autoprefixer [as seen here](https://tailwindcss.com/docs/guides/gatsby), so I don't think we should remove it 😃

olivermrbl: My bad. Added again :)

pKorsholm:
lgtm

pKorsholm (commented):
LGTM! 😄

PR #218, opened 2022-01-04
Feat: Revamp input field

pKorsholm:
**What** 
- input fields for text and number types
- Wrapper and header fundamentals for input to be used with later input types (such as dropdown)

**How**
- use specialised classes in `global.css` for different browsers which disable the number spinners on number inputs, classname `remove-number-spinner` which is applied to input tags to disable spinners locally but not globally
- Using refs for the input tag along with `useImperativeHandle` to properly use the incoming ref from `React.forwardRef`
- custom handlers which using references to the input field for native increment and decrement handling
- using `e.preventDefault()` on the `InputContainer` to prevent native mousedown handling (which removes focus and styling from the input)
  - `stopPropagation()` is used on the input field which ensures that clicks in the field are handled according to the native click handler of the field


src/components/fundamentals/input-header.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+const InputHeader: React.FC<InputHeaderProps> = ({
+  label,
+  required = false,
+  withTooltip = false,
+  tooltipText,
+  tooltipProps,
+}) => {
+  return (
+    <div className="w-full flex pb-1.5 font-semibold text-small leading-xsmall text-grey-50 items-baseline">
```
kasperkristensen, line 20: In the global.css file there is a number of text utility classes that you can use instead of writing out the all the classes. In this instance you could add the class `inter-small-semibold`. This would also set the line height to 20px and not 16px as you have done here which I believe is wrong according to the Figma? (Please correct me if I'm wrong)

src/components/fundamentals/input-header.tsx
```diff
@@ -0,0 +1,32 @@
...
+      {label}
+      {required && <div className="text-red-600 "> *</div>}
```
kasperkristensen, line 22: In the figma the color rose-50 is used for warnings/reds so I think we should also use this here 😁

src/components/fundamentals/input-header.tsx
pKorsholm: you are right 😄  had the right font size but not line height, it's fixed

src/components/molecules/input/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+      invalid,
+      placeholder,
+      defaultValue,
+      step,
+      min,
+      max,
+      inline,
+      label,
+      name,
+      type,
```
zakariaelas, line 52: nit: curious to know why the props do not extend something like `InputHTMLAttributes<HTMLInputElement>`? This would allow the consumer of this component to pass in any html input attribute. We could use a rest parameter and pass it to the native `<input>` element. 

_PS: not 100% sure if the right type to extend is exactly `InputHTMLAttributes<HTMLInputElement>`_

src/components/molecules/input/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+      inputStyle,
```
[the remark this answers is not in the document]
pKorsholm: they are not, I have removed them so errors will show when you try to use the input with invalid parameters

src/components/molecules/input/index.tsx
pKorsholm: You are absolutely right, i have now made the props as so: 
```javascript 
type InputProps = React.InputHTMLAttributes<HTMLInputElement> & {
  label: string
  deletable?: boolean
  key?: string
  onDelete?: MouseEventHandler<HTMLSpanElement>
  onChange?: ChangeEventHandler<HTMLInputElement>
  onFocus?: FocusEventHandler<HTMLInputElement>
  withTooltip?: boolean
  tooltipText?: string
  tooltipProps?: any
  props?: React.HTMLAttributes<HTMLDivElement>
}
```

Where only the additional properties are added. Furthermore i then spread the rest of the props in the component and apply them to the field so we can pass all values existing on an `<input />`: 

```javascript
const InputField = React.forwardRef(
  (
    {
      placeholder,
      ...
      tooltipProps = {},
      props,
      ...fieldProps
    }: InputProps,
    ref
  ) => {
  ...
  return (
    ...
    <input
      ...
      {...fieldProps}
    />
    ...
  )
}
```

zakariaelas (approved):
LGTM w. small nit :+1:

src/components/fundamentals/input-header.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+const InputHeader: React.FC<InputHeaderProps> = ({
+  label,
+  required = false,
+  withTooltip = false,
+  tooltipText,
+  tooltipProps,
+}) => {
+  return (
+    <div className="w-full flex inter-small-semibold text-grey-50 items-baseline">
```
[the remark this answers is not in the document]
olivermrbl: I actually don't think we have design for tooltips yet, but correct me If I am wrong @pKorsholm. Could not find them in Figma.

So we might need to do some adjustments when the design is ready and leave them for now :)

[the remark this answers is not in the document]
pKorsholm: Yeah you are right, I have centered it for now, if it's required to change we just change the tooltip component and the layout will switch :)

Issue #156, opened 2021-10-18
Able to click next page when no more options are to be shown


sebastiannicolajsen:
On `/a/swaps` and `/a/returns` you are able to click next to show more swaps / returns even though no more are to be shown:
![Skærmbillede 2021-10-18 kl  15 46 53](https://user-images.githubusercontent.com/80953876/137743996-f1460ade-e674-498b-9f4b-2db1bba4d73a.png)
![Skærmbillede 2021-10-18 kl  15 46 56](https://user-images.githubusercontent.com/80953876/137744009-edae0484-bc01-49af-a962-95307fd5becb.png)

PR #219, opened 2022-01-04
Fix/swaps and returns pages

NayeBeckham:
Solves [#156](https://github.com/medusajs/admin/issues/156)

**What's the problem**
On `/a/swaps` and `/a/returns` you were able to click more pages even when there wasn't any data.

**Why we should fix it**
It looks better if the application shows the necessary pages.

**How I fixed it**
Added a disabled option on the next button when the length of the data is greater than the limit. Like:
`const nextReturn = returns && returns.length >= limit`
In the button:
`disabled={!nextReturn}`

Screen Shot 2022-01-04 at 13 59 18

Screen Shot 2022-01-04 at 13 59 24

**Testing**
No needed.

Hope it is useful, tell me anything, thank you!

### DIFF
--- a/src/domain/orders/returns/index.js
+++ b/src/domain/orders/returns/index.js
@@ -130,6 +130,8 @@ const ReturnIndex = ({}) => {
     })
   }
 
+  const nextReturn = returns && returns.length >= limit
+
   return (
     <Flex flexDirection="column" pb={5} pt={5}>
       <Flex>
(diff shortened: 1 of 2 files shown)


olivermrbl:
Thanks for the contribution @NayeBeckham ! As part of the new admin version, we've decided to remove the swaps and returns pages and instead offer extensive filtering capabilities. So closing this one down

See #325

PR #220, opened 2022-01-05
fix: attributes on icon type

kasperkristensen:
Fixes a bad implementation idea by me. Changes IconType to be IconType { .. } & React.SVGAttributes<SVGElement>, so we actually get code suggestions.


olivermrbl (approved):
LGTM

PR #221, opened 2022-01-05
feat: revamp two split body

kasperkristensen:
**What**
- Adds two split view

**How**
To use the component you will add a `<TwoSplitPane></TwoSplitPane>` which can hold two `BodyCard` components as children. If you add more then two children the component will throw an error.

**Responsiveness**
I've added some very rudimental responsiveness to the component, but I definitely think we should revisit this once Ludvig has taken a look at how this component should look on different screen sizes. 

**Note**
PR `feat/revamp-buttons` and `fix/icon-type` should be merged beforehand

Skærmbillede 2022-01-05 kl  17 23 32


src/components/molecules/actionables.tsx
```diff
@@ -0,0 +1,71 @@
...
+type ActionablesProps = {
+  actions: ActionType[]
+}
+
+/**
+ * A component that accepts multiple actionables and renders them as a dropdown menu.
+ * If only a single actionable is provided, it will render a button instead.
+ */
+const Actionables: React.FC<ActionablesProps> = ({ actions }) => {
+  if (actions.length === 0) {
```
olivermrbl, line 23: Should check if actions is `undefined`, since they are not required
```suggestion
  if (actions?.length === 0) {
```

src/components/molecules/actionables.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+export type ActionType = {
+  label: string
+  onClick: (e: React.MouseEvent<HTMLButtonElement>) => void
+  variant?: "normal" | "danger"
+  icon: React.ReactNode
+}
+
+type ActionablesProps = {
+  actions: ActionType[]
```
olivermrbl, line 15: Should be optional right?

src/components/organisms/body-card.tsx
```diff
@@ -0,0 +1,70 @@
...
+const BodyCard: React.FC<BodyCardProps> = ({
+  title,
+  subtitle,
+  events,
+  actionables,
+  children,
+}) => {
+  const { isScrolled, scrollListener } = useScroll()
+  return (
+    <div className="rounded-rounded border bg-grey-0 border-grey-20 h-full overflow-hidden flex flex-col min-h-[500px] w-full medium:w-1/2 relative">
```
olivermrbl, line 24: Should have a smaller min-height. See currency settings [here](https://www.figma.com/file/dlraOnQvQnrHQL0A7Ysbrs/Webapp-2.0?node-id=74%3A87).

olivermrbl (changes requested):
Great work Kasper! Minor changes required 👍

src/components/organisms/body-card.tsx
olivermrbl: Suggestion would be `400px`

pKorsholm:
Should the `BodyCard` be able to be used without putting it in a `TwoSplitPane`?

If you try that now it's still half split, I would suggest we make it full width and have the `TwoSplitPane` adjust the width of the `BodyCard`. Then in sub menus with only one side we would still use the `TwoSplitPane` and have the same halving effect, but we can use the `BodyCard` for full width components.

I.e remove `medium:w-1/2` from the outer `div` of `BodyCard` and adjust the `TwoSplitPane` to map children like so: 

```javascript
{Children.map(children, (child, i) => {
  return (
    <div className="medium:w-1/2" key={i}>
      {child}
    </div>
  )
})}
```

kasperkristensen:
> Should the `BodyCard` be able to be used without putting it in a `TwoSplitPane`?
> 
> If you try that now it's still half split, I would suggest we make it full width and have the `TwoSplitPane` adjust the width of the `BodyCard`. Then in sub menus with only one side we would still use the `TwoSplitPane` and have the same halving effect, but we can use the `BodyCard` for full width components.
> 
> I.e remove `medium:w-1/2` from the outer `div` of `BodyCard` and adjust the `TwoSplitPane` to map children like so:
> 
> ```js
> {Children.map(children, (child, i) => {
>   return (
>     <div className="medium:w-1/2" key={i}>
>       {child}
>     </div>
>   )
> })}
> ```

Totally agree! I've added it to the update 👍

olivermrbl (approved):
LGTM!

PR #222, opened 2022-01-05
fix: update storybook

kasperkristensen:
**What**

- Updates storybook, and its settings to allow it to be used with TailwindCSS.
- Adds stories to atoms, fundamentals, some organisms and some molecules


package.json
```diff
@@ -5,15 +5,16 @@
...
+    "postcss": "^8.4.5"
   },
   "dependencies": {
     "@emotion/core": "^10.0.28",
     "@emotion/styled": "^10.0.27",
     "@popperjs/core": "^2.10.2",
     "@radix-ui/react-dialog": "^0.1.5",
     "@rebass/forms": "^4.0.6",
-    "@storybook/theming": "^5.3.19",
+    "@storybook/theming": "^6.4.9",
```
pKorsholm, line 17: should this be a dev-dependency?

pKorsholm (approved):
LGTM! 

it's gonna be great having storybook as part of the development flow, very nice! 😄

package.json
kasperkristensen: Yea, I think it should be. Will change it in a sec 😄

PR #223, opened 2022-01-06
Feat: Revamp searchable dropdown

pKorsholm:
**What**
- Select component for admin revamp including:
  - multiselect/single select functionality
  - optional search functionality
  - optional 'select all` functionality

**How**
- using the existing component import `react-multi-select-component` used in the existing multiselect
  - styling that component 
    - creating custom itemviewer 
    - overriding existing css classes
- specifically changing these two classes
  - src/components/molecules/multi-select/index.tsx
  - src/assets/styles/global.css

**Properties**
The `<Select />` component by default is a standard dropdown component styled to the admin revamp layout. It requires the following properties: 
- `label:string` - label to display
- `options: { label: string; value: string; disabled?: boolean }[]` - options to chose from (can be individually disabled)
- `value: { label: string; value: string }[] ` - the value of the select, this is an array with length 1 when used as a regular dropdown
- `onChange: (values: any[]) => void` - handler for changes in selection

The component also has the following optional components for controlling behaviour:
- `hasSelectAll?: boolean` - adds a "select all" element as the first child of the dropdown
- `enableSearch?: boolean` - enables a search field using fuzzy search at the top of the list
- `isMultiSelect?: boolean` - enables multi select functionality
 
along with the rest of the properties for the [`MultiSelect` component](https://github.com/hc-oss/react-multi-select-component#-props)

*dependent on merging `input`*


kasperkristensen (commented):
Looks great! But it seems like it does not handle selecting multiple values, when ever you select a new value it removes the ones previously selected. And if I enable `hasSelectAll` then that option does not seem to work either.

src/assets/svg-2.0/search.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="20" height="20" viewBox="0 0 20 20" fill="none" xmlns="http://www.w3.org/2000/svg">
```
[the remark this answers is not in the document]
pKorsholm: not exactly, it's used as an svg in css to style the search field like: 
```css
.select-panel > div::before {
  content: url("../../assets/svg-2.0/search.svg");
  width: 2rem;
  ...
}
```

src/components/molecules/select/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+          overrideStrings={{ search: "Search...", ...overrideStrings }}
+          ClearIcon={
+            <span className="text-grey-40">
+              <XCircleIcon size={20} />
+            </span>
+          }
+          onChange={handleSelect}
+          {...selectOptions}
+          disableSearch={!enableSearch}
+        />
```
zakariaelas, line 124: Although the clearIcon is being rendered when the input is searchable, I think it'd be cool to have it or any way of deleting/resetting the multiselect after selection

image

src/components/molecules/select/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+  value: { label: string; value: string }[]
+  hasSelectAll?: boolean
+  isLoading?: boolean
+  shouldToggleOnHover?: boolean
+  overrideStrings?: object
+  onChange: (values: any[]) => void
+  disabled?: boolean
+  enableSearch?: boolean
+  isCreatable?: boolean
+  onCreateOption: any
```
zakariaelas, line 38: perhaps optional?

zakariaelas (approved):
Really awesome Philip :fire: . LGTM w. small comments

src/components/molecules/select/index.tsx
pKorsholm: done 👍

PR #226, opened 2022-01-10
Feat: Revamp avatar

pKorsholm:
**What** 
- Implement the avatar atom using radix avatar component

**How**
- adapt existing styling to the radix avatar component
- make avatar scale to 100% width and height to fit whatever contain it's put in

**Why**
- To avoid having multiple avatar components
- to utilise the fallback function of radix avatar


olivermrbl (approved):
LGTM 🚀

PR #228, opened 2022-01-10
feat: Revamp currency settings

olivermrbl:
**What**
- Currency settings page

_note: this pr is dependent on #221_


src/assets/styles/global.css
```diff
@@ -164,11 +164,11 @@
     @apply p-0 -mx-3 mt-1 outline-0 cursor-text;
 
     .dropdown-container {
-      @apply border-none bg-inherit focus-within:shadow-none cursor-text !important;
+      @apply border-none bg-inherit relative focus-within:shadow-none cursor-text !important;
     }
 
     .dropdown-content {
-      @apply rounded-rounded w-full -mt-3 pt-0 -top-full shadow-input !important;
+      @apply rounded-rounded w-full -mt-3 pt-0 -top-full shadow-input absolute !important;
```
olivermrbl, line 171: To make dropdown overflow parent container

src/domain/settings/currencies.js
```diff
@@ -1,168 +1,116 @@
...
+      return
     }
-  })
+
+    setCurrenciesOnLoad()
+  }, [store, isLoading])
 
   const handleChange = currencies => {
-    setCurrencies(currencies)
+    setStoreCurrencies(currencies)
```
pKorsholm, line 51: Dont know what the expected behaviour is here, but right now you can remove the default store currency from the store-currencies and save the resulting currency configuration:

image

We should probably enforce the behvaiour both here and in the core (i have created a ticket)

src/domain/settings/currencies.js
```diff
@@ -1,168 +1,116 @@
...
   }
 
-  const onSubmit = data => {
+  const onSubmit = async e => {
+    e.preventDefault()
     try {
-      update({
-        default_currency_code: data.default_currency_code,
-        currencies: selectedCurrencies.map(c => c.value),
+      await update({
```
pKorsholm, line 57: do we use then chaining or try catch (for consistency)?

pKorsholm (approved):
LGTM

I think it's fine not fixing overflow for now, and doing that in a later pr, for now #234 looks to change up bodycard and twosplit quite a bit

pKorsholm:
Also `currencies.js` should be converted to `currencies.tsx`

src/domain/settings/currencies.js
olivermrbl: Should be fixed as part of core [PR](https://github.com/medusajs/medusa/pull/984) 👍

PR #229, opened 2022-01-11
Feat: Revamp delete prompt

pKorsholm:
**What**
- add a danger button to button component and `global.css`
- fix modal widths
- Add delete prompt 

Delete promp has the following properties: 
- `heading: string`: heading for the prompt ie. 'Remove user'
- `text: string`: text for prompt body ie. 'Are you sure you want to remove this user?'
- `successText?: string`: success text for toaster
- `handleClose: () => void`: close function for modal
- `onDelete: () => Promise<void>`: function which deletes the entity in question

image

### DIFF
--- a/src/components/organisms/delete-prompt.tsx
+++ b/src/components/organisms/delete-prompt.tsx
@@ -5,17 +5,21 @@ import useMedusa from "../../hooks/use-medusa"
 import { getErrorMessage } from "../../utils/error-messages"
 
 type DeletePromptProps = {
-  heading: string
-  text: string
+  heading?: string
+  text?: string
   successText?: string
+  cancelText?: string
+  confirmText?: string
   handleClose: () => void
   onDelete: () => Promise<void>
 }
 
 const DeletePrompt: React.FC<DeletePromptProps> = ({
-  heading,
-  text,
-  successText,
+  heading = "Are you sure?",
+  text = "Are you sure you want to delete?",
+  successText = "Delete successful",
+  cancelText = "No, cancel",
+  confirmText = "Yes, remove",
   handleClose,
   onDelete,
 }) => {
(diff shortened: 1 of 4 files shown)


src/components/organisms/delete-prompt.tsx
```diff
@@ -0,0 +1,77 @@
...
+            >
+              No, cancel
+            </Button>
+            <Button
+              loading={isLoading}
+              size="small"
+              className="w-24 text-small justify-center"
+              variant="danger"
+              onClick={handleSubmit}
+            >
```
olivermrbl, line 67: nit: Add `confirmText = "Yes, remove"` and `cancelText="No, cancel"` to props. I know this might not be used much, but we might as well offer the ability to do so.

src/components/organisms/delete-prompt.tsx
```diff
@@ -0,0 +1,77 @@
...
+  return (
+    <Modal handleClose={handleClose}>
+      <Modal.Body>
+        <Modal.Content>
+          <div className="flex flex-col">
+            <span className="inter-large-semibold">
+              {heading || "Are you sure?"}
+            </span>
+            <span className="inter-base-regular mt-1 text-grey-50">
+              {text || "Are you sure you want to delete?"}
```
olivermrbl, line 47: very, very nit: Set default value in props instead. Goes for all props.

src/components/organisms/delete-prompt.tsx
```diff
@@ -0,0 +1,77 @@
...
+}) => {
+  const { toaster } = useMedusa("store")
+  const [isLoading, setIsLoading] = useState(false)
+
+  const handleSubmit = e => {
+    e.preventDefault()
+
+    setIsLoading(true)
+    onDelete()
+      .then(() => toaster(successText || "Delete successful", "success"))
```
olivermrbl, line 30: very, very nit: Same as below.

olivermrbl (approved):
LGTM, very nice!

Couple of comments :)

PR #234, opened 2022-01-12
fix: two split pane + layout

zakariaelas:
### Fixes two split layout

Both children should have the same height, or in other words, the smaller sibling should follow the larger sibling in terms of height

**Before**
Right pane's height is not equal to left pane's height

149199153-40256338-bcbb-4ac0-bd2c-090ffe8092c3

**After**
Right pane's height is equal to left pane's height

149199112-a3a648f4-760a-4e5c-85b8-082318dc6c38

### Fixes layout

When the content grows beyond a certain height, it overflows the bottom padding initially set by the layout

**Before**
There is no bottom padding when the page is scrolled all the way

149198692-1f9a41bd-3b95-49d2-a605-59b0213fed5e

**After**

149198600-14810719-95ca-4331-9c3c-8a2fed2165b5(1)

## Why 

- ensures consistency

## How

- **two split pane**: use grid layout as opposed to flex layout. When screen size > medium, we use a 2-column layout, otherwise, we only use one column. 
- **layout**: 
   - remove `height: 100%` from our Layout component's `main` tag and use the default `height: auto` so that it grows with the content.
   - replace `overflow-y: scroll` by `overflow-y: auto` in our `Layout` component. Not entirely sure if it's okay to do that, so let me know what you think! 

Let me know what you guys think as I suspect you might have solved them differently in your on-going tasks @kasperkristensen @pKorsholm @olivermrbl :+1:

### DIFF
--- a/src/components/templates/layout.tsx
+++ b/src/components/templates/layout.tsx
@@ -8,8 +8,8 @@ const Layout: React.FC = ({ children }) => {
       <Sidebar />
       <div className="flex flex-col flex-1">
         <Topbar />
-        <div className="large:px-xlarge py-xlarge bg-grey-5 min-h-content overflow-y-scroll">
-          <main className="xsmall:mx-base small:mx-xlarge medium:mx-4xlarge large:mx-auto large:max-w-7xl large:w-full h-full">
+        <div className="large:px-xlarge py-xlarge bg-grey-5 min-h-content overflow-y-auto">
+          <main className="xsmall:mx-base small:mx-xlarge medium:mx-4xlarge large:mx-auto large:max-w-7xl large:w-full">
             {children}
           </main>
         </div>
--- a/src/components/templates/layout.tsx
+++ b/src/components/templates/layout.tsx
@@ -9,7 +9,7 @@ const Layout: React.FC = ({ children }) => {
       <div className="flex flex-col flex-1">
         <Topbar />
         <div className="large:px-xlarge py-xlarge bg-grey-5 min-h-content overflow-y-auto">
-          <main className="xsmall:mx-base small:mx-xlarge medium:mx-4xlarge large:mx-auto large:max-w-7xl large:w-full">
+          <main className="xsmall:mx-base small:mx-xlarge medium:mx-4xlarge large:mx-auto large:max-w-7xl large:w-full h-full">
             {children}
           </main>
         </div>
--- a/src/components/organisms/body-card.tsx
+++ b/src/components/organisms/body-card.tsx
@@ -1,3 +1,4 @@
+import clsx from "clsx"
 import React from "react"
 import { useScroll } from "../../hooks/use-scroll"
 import Button from "../fundamentals/button"
@@ -11,18 +12,26 @@ type BodyCardProps = {
     onClick: (e: React.MouseEvent<HTMLButtonElement>) => void
   }[]
   actionables?: ActionType[]
-}
+} & React.HTMLAttributes<HTMLDivElement>
 
 const BodyCard: React.FC<BodyCardProps> = ({
   title,
   subtitle,
   events,
   actionables,
+  className,
   children,
+  ...rest
 }) => {
   const { isScrolled, scrollListener } = useScroll({ threshold: 16 })
   return (
-    <div className="rounded-rounded border bg-grey-0 border-grey-20 h-full overflow-hidden flex flex-col min-h-[350px] w-full relative">
+    <div
+      className={clsx(
+        "rounded-rounded border bg-grey-0 border-grey-20 h-full overflow-hidden flex flex-col min-h-[350px] w-full relative",
+        className
+      )}
+      {...rest}
+    >
       {isScrolled && (
         <div className="absolute top-0 left-0 right-0 bg-gradient-to-b from-grey-0 to-transparent h-xlarge z-10" />
       )}
--- a/src/components/templates/two-split-pane.tsx
+++ b/src/components/templates/two-split-pane.tsx
@@ -1,14 +1,26 @@
 import React, { Children } from "react"
+import { useComputedHeight } from "../../hooks/use-computed-height"
 
 const TwoSplitPane: React.FC = ({ children }) => {
   const childrenCount = Children.count(children)
+  const { ref, height } = useComputedHeight(32)
+
+  const heightClass = height
+    ? {
+        gridTemplateRows: `${height}px`,
+      }
+    : null
 
   if (childrenCount > 2) {
     throw new Error("TwoSplitPane can only have two or less children")
   }
 
   return (
-    <div className="grid gap-large grid-cols-1 medium:grid-cols-2">
+    <div
+      className="grid gap-large grid-cols-1 medium:grid-cols-2"
+      style={heightClass}
+      ref={ref}
+    >
       {Children.map(children, (child, i) => {
         return (
           <div className="w-full h-full" key={i}>
--- a/src/components/templates/two-split-pane.tsx
+++ b/src/components/templates/two-split-pane.tsx
@@ -8,10 +8,10 @@ const TwoSplitPane: React.FC = ({ children }) => {
   }
 
   return (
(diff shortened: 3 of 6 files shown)


kasperkristensen:
I've also been trying to fix this myself but haven't found a good solution yet. I think this is really close to being perfect. But it is still not completely inline with the the intended design. If we go by Ludvigs design then the TwoSplitPane and it's children should not be able to grow larger then the screen, and in case of any overflow the BodyCard components should scroll as seen [here](https://www.figma.com/file/dlraOnQvQnrHQL0A7Ysbrs/Webapp-2.0?node-id=64%3A289). I got it to "sort of" work by adding height values to the two <Router/> components higher in the component tree, but the look of it was not very consistent on every page, as this limitation should not apply to any page that is not wrapped in a TwoSplitPane. Do you have any good ideas on how we can get this to work in @zakariaelas? 😄

zakariaelas:
Very true! Thanks for pointing me to those designs, completely missed them :sweat_smile:! I personally don't know of a bulletproof way to do this, but here are my 2 cents:

One approach, just like the one you tried, is to set `height:100%` a bit up in the component tree. The disadvantage with that is that you have to keep setting `height: 100%` to all the children down the component tree until you reach the top-level parent of the split layout, and even then, it would require some fiddling to set the right height for the grid row (containing the spit layout cards side by side). 

Another approach would be to set some fixed height to the grid row. I think this corresponds to Ludvig's designs where the cards keep the same height regardless of whether the content overflows or not. To set a fixed height to the grid row, we can use `grid-template-rows: calc(100vh - 156px)`. The trick is to use the entire viewport height as a baseline and just subtract all the vertical padding, margins, and content surrounding our row (see image below).

image

image

You could argue that it is a bit unfortunate that the `156px` is hardcoded, but we could probably improve this a little further if we use css variables to capture the surroundings or maybe even some js (ugliness):

```jsx
const TwoSplitPane: React.FC = ({ children }) => {
  const childrenCount = Children.count(children)
  const gridRef = useRef(null)
  const gridRowHeight = useRef(0)
  const heightClass = gridRowHeight.current ? "medium:grid-rows-[" + gridRowHeight.current + "px]" : ""
  
  useLayoutEffect(() => {
    if(gridRef.current) {
      let { top } = gridRef.current.getBoundingClientRect()
      let height = window.innerHeight
      // take the inner height of the window, subtract 32 from it (for the bottom padding), then subtract that from the top position of our grid row (wherever that is)
      gridRowHeight.current = height - 32 - top
    }
  }, [])

  if (childrenCount > 2) {
    throw new Error("TwoSplitPane can only have two or less children")
  }

  return (
    <div ref={gridRef} className={clsx("grid gap-large grid-cols-1 medium:grid-cols-2 ", heightClass)}>
      {Children.map(children, (child, i) => {
        return (
          <div className="w-full h-full" key={i}>
            {child}
          </div>
        )
      })}
    </div>
  )
}
```
A visual representation of how the grid row height is calculated:

image


However, all in all, just like you mentioned, because this layout is only desired for the `TwoSplit` component, whatever the best approach may be, I think it's much better if it only affects elements under this component so as not to have any conflicting issues when working on other layouts.

What do you guys think? Do any other approaches come to mind? Also, do you prefer the hard coded value or the more flexible, yet ugly, alternative which uses JS?

Thanks for reading :smile:

pKorsholm:
playing around with this for at bit I personally like the ugly computed js version better because of the flexibility in sizing. However, unrelated to the two-split-pane, but relevant to what you previously suggested as a solution I think we need to set the height of the two router components higher in the component tree as well to accomodate layouts like team settings which utilize the full screen height 

image

Which is not possible if we dont allow the router components to extend to 100% height, unless we hardcode or calculate a height in the team component as we would do in the two-split-pane now.

src/components/templates/layout.tsx
```diff
@@ -8,8 +8,8 @@ const Layout: React.FC = ({ children }) => {
       <Sidebar />
       <div className="flex flex-col flex-1">
         <Topbar />
-        <div className="large:px-xlarge py-xlarge bg-grey-5 min-h-content overflow-y-scroll">
-          <main className="xsmall:mx-base small:mx-xlarge medium:mx-4xlarge large:mx-auto large:max-w-7xl large:w-full h-full">
+        <div className="large:px-xlarge py-xlarge bg-grey-5 min-h-content overflow-y-auto">
+          <main className="xsmall:mx-base small:mx-xlarge medium:mx-4xlarge large:mx-auto large:max-w-7xl large:w-full">
```
pKorsholm, line 12: I think this should still have h-full to allow full height children with negative space in components

kasperkristensen:
I'd argue we should go with the JS solution even if it is a little "ugly". One problem with it though is that you can't use string concats for Tailwind classes. Further more as I talked with Oli about earlier, we will need the same ability to limit a component to the remaining screen size. Therefore I suggest we do something like this:

```js

// use-computed-height.ts
import { useLayoutEffect, useRef } from "react"

export const useComputedHeight = (bottomPad: number) => {
  const ref = useRef(null)
  const heightRef = useRef(0)

  useLayoutEffect(() => {
    if (ref.current) {
      let { top } = ref.current.getBoundingClientRect()
      let height = window.innerHeight
      heightRef.current = height - bottomPad - top
    }
  }, [bottomPad])

  return { ref, height: heightRef.current }
}
```

which we can then use as:

```js
import React, { Children } from "react"
import { useComputedHeight } from "../../hooks/use-computed-height"

const TwoSplitPane: React.FC = ({ children }) => {
  const childrenCount = Children.count(children)
  const { ref, height } = useComputedHeight(32)

  const heightClass = height
    ? {
        gridTemplateRows: `${height}px`,
      }
    : null

  if (childrenCount > 2) {
    throw new Error("TwoSplitPane can only have two or less children")
  }

  return (
    <div
      className="grid gap-large grid-cols-1 medium:grid-cols-2"
      style={heightClass}
      ref={ref}
    >
      {Children.map(children, (child, i) => {
        return (
          <div className="w-full h-full" key={i}>
            {child}
          </div>
        )
      })}
    </div>
  )
}

export default TwoSplitPane
```

Let me know what you think, if you agree then I will push it to this branch 😄

zakariaelas:
@pKorsholm you raised a valid point :+1:. Will update some of the components with `h-full` later today

@kasperkristensen love the hooks implementation :heart_eyes: , really really neat and makes things less "ugly" and reusable. Go ahead and push this in :+1:

olivermrbl:
Tried it out @kasperkristensen and it seems to be working really well - great job! 

One thing to be aware of is, that with this implementation we should no longer place "single panes" inside the two-split pane component (don't know if we ever have done this anyhow, but now it's obvious 😅 ) because this makes them expand full height. This is due to the fact that the two-split panes are always full height now.

Though we currently don't have any screens in Figma with two-split panes not spanning full height, I think this solution is as perfect as it gets right now.  

I say, let's get this in and move on. Let me know, what you guys think 😃

olivermrbl:
As a result of our discussion Friday, I've added a hook to`useComputedHeight` to receive height updates on resize events. 

I've come to think of the cases where we have a "single pane" (e.g. Personal settings), and we don't want that to be of full height, but instead have it fit its content until the content expands the height of the pane and then let the two-split scrollable behavior kick in. Hope it makes sense 😅 

I've opened a PR [here](https://github.com/medusajs/admin/pull/242) for a potential solution (very simple). 

Let me know, what you guys think @zakariaelas @pKorsholm @kasperkristensen ?

pKorsholm:
I agree that we should get @olivermrbl's pr in and get this merged, after all we are only merging to a develop branch and the faster we begin to use the solution the faster we find the potential limits and can address those 😄  lets get it in asap

pKorsholm:
> @pKorsholm maybe just add the h-full to the layout that you were talking about in your comments to #238? 🚀

done 😄

olivermrbl:
Should we get this merged in now? 😊 LGTM

zakariaelas:
@olivermrbl all good on my side, we can always iterate back on this if we need to :+1:

PR #237, opened 2022-01-13
suggested fix for issue 235

dsheyp:
-  inventory_quantity: cast string values to int
- allow 0 as a value for inventory_quantity


agustif:
Hey I also hit this one and this PR fixes it for me too, only way I could update my products via admin besides editing directly in database

olivermrbl:
Thanks! Solved in #325

PR #238, opened 2022-01-14
feat: revamp ui regions

kasperkristensen:
**What**
- Revamps the Region section of settings
- Implements new Radio Group component
- Implements new Shipping Option component

**Note**
Merge #234 beforehand as this has been merged into this branch

image
image
image
image
image


src/components/templates/layout.tsx
```diff
@@ -8,8 +8,8 @@ const Layout: React.FC = ({ children }) => {
       <Sidebar />
       <div className="flex flex-col flex-1">
         <Topbar />
-        <div className="large:px-xlarge py-xlarge bg-grey-5 min-h-content overflow-y-scroll">
-          <main className="xsmall:mx-base small:mx-xlarge medium:mx-4xlarge large:mx-auto large:max-w-7xl large:w-full h-full">
+        <div className="large:px-xlarge py-xlarge bg-grey-5 min-h-content overflow-y-auto">
+          <main className="xsmall:mx-base small:mx-xlarge medium:mx-4xlarge large:mx-auto large:max-w-7xl large:w-full">
```
pKorsholm, line 12: I think this needs to be h-full acc. to the discussion in #234

src/domain/settings/regions/edit-shipping.tsx
```diff
@@ -0,0 +1,248 @@
...
+                size="small"
+                className="w-[127px] justify-center"
+              >
+                Cancel Changes
+              </Button>
+              <Button
+                type="submit"
+                variant="primary"
+                size="small"
+                className="w-[127px] justify-center"
```
pKorsholm, line 236: This width should be added to tailwind-config or we should use `w-32` and accept that it's 128px instead

src/domain/settings/regions/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+                    r.payment_providers
+                      .map(pp => paymentProvidersMapper(pp.id).label)
+                      .join(", ") || "not configured"
+                  } - Fulfillment providers: ${
+                    r.fulfillment_providers
+                      .map(fp => fulfillmentProvidersMapper(fp.id).label)
+                      .join(", ") || "not configured"
+                  }`
+                  return (
+                    <RadioGroup.Item
```
pKorsholm, line 90: can we make the entire radiogroup item clickable instead of just the radio button? it seems a bit un-intuitive

src/domain/settings/regions/new-shipping.tsx
```diff
@@ -0,0 +1,298 @@
...
+  } = useAdminShippingProfiles()
+  const [adminOnly, setAdminOnly] = useState(false)
+  const [options, setOptions] = useState([])
+  const [selectedOption, setSelectedOption] = useState(null)
+  const [profileOptions, setProfileOptions] = useState([])
+  const [selectedProfile, setSelectedProfile] = useState(null)
+  const createShippingOption = useAdminCreateShippingOption()
+  const toaster = useToaster()
+
+  console.log(errors)
```
pKorsholm, line 36:
```suggestion

```

src/domain/settings/regions/new-shipping.tsx
```diff
@@ -0,0 +1,298 @@
...
+      onSuccess: () => {
+        toaster("Successfully created shipping option", "success")
+        if (onCreated) {
+          onCreated()
+        }
+        onClick()
+      },
+      onError: error => {
+        toaster(getErrorMessage(error), "error")
+        console.error(error)
```
pKorsholm, line 96:
```suggestion

```

src/domain/settings/regions/new.tsx
```diff
@@ -0,0 +1,228 @@
...
+    )
+  }
+
+  const onSave = data => {
+    if (!data.countries?.length) {
+      toaster("Choose at least one country", "error")
+      return
+    }
+
+    console.log(data.currency_code)
```
pKorsholm, line 84:
```suggestion

```

src/domain/settings/regions/edit-shipping.tsx
kasperkristensen: Generally don't think we should use the built in spacings from Tailwind such as w-32, as we have implemented our own spacings according to Ludvig's design. I've added a w-eventButton in this case 😄

src/components/templates/layout.tsx
kasperkristensen: I think it is better to add this to the fix/two-split-pane PR instead, and not as part of this PR 😄

src/domain/settings/regions/edit-shipping.tsx
pKorsholm: perfect 😄

pKorsholm (approved):
lgtm

src/domain/settings/regions/details.tsx
```diff
@@ -0,0 +1,343 @@
...
+
+  useEffect(() => {
+    if (paymentIsLoading) return
+    setPaymentOptions(payment_providers.map(c => paymentProvidersMapper(c.id)))
+  }, [payment_providers, paymentIsLoading])
+
+  useEffect(() => {
+    if (fulfillmentIsLoading) return
+    setFulfillmentOptions(
+      fulfillment_options.map(c => fulfillmentProvidersMapper(c.provider_id))
```
zakariaelas, line 74: this throws when I run it locally, probably because i'm using the tekla db so there might be an isssue with that. However, is `useAdminFulfillmentOptions` fails for whatever reason, this will throw because `fulfillment_options` will be undefined, so maybe add a tiny `?`. 

Note that if you use `?`, `fulfillmentOptions` (the local state) will be undefined so maybe default to an empty array or use a plain old boring `if (fulfillment_options) setFulfillmentOptions(...)`

zakariaelas (commented):
LGTM w. few comments :+1: Great work!

PR #239, opened 2022-01-16
feat: revamp store settings

zakariaelas:
### What

- store settings
- adds a `type` property to the `BodyCard` component props type

### Why

- if the `BodyCard` is inside a form, then all events (buttons) will behave as `type="submit"` buttons. Instead, consumer of `BodyCard` should be able to alter this behavior as needed


src/domain/settings/details.tsx
```diff
@@ -0,0 +1,128 @@
...
+            label="Store name"
+            name="name"
+            placeholder="Medusa Store"
+            ref={register}
+          />
+
+          <h6 className="mt-2xlarge inter-base-semibold">Advanced settings</h6>
+          <span className="inter-small-regular text-grey-50">
+            Manual-fulfillment via manual
+          </span>
```
olivermrbl, line 87: Think this is a design flaw. Seems rather misplaced 😅 I'll ping Ludvig but let's remove it for now.

olivermrbl (approved):
LGTM @zakariaelas 🙌 ! 

Only a tiny change. Feel free to merge after 👍

olivermrbl:
I've opened a [PR](#242), that would allow you to put the `<BodyCard />` inside `<TwoSplitPane />` and achieve the same scrollable behavior when content expands the view height as seen on attached below:

image

Would love to hear what you think of that approach @zakariaelas 😊

PR #240, opened 2022-01-16
feat: revamp ui return reasons

zakariaelas:
### What 

- return reasons

Not entirely sure how the duplication is supposed to behave in this case, but for now, when a user clicks "Duplicate Reason", the same modal that is used for creating a return reason is shown with prefilled values from the duplicated reason. The user can modify any field then proceed to create the return reason. 

One other way to do it is to just append the return reason to the list of return reasons, and then select the return reason so that the user can change the fields on the right panel, however, the `value` field of return reasons is immutable, so that'll make the `ReturnReasonDetail` a tiny bit more complex, so that's why I went with the first approach.

_Note: The `RadioGroup` component is the same as in PR #238. Once PR #238 is merged, I'll make sure to merge the component to this branch_


src/components/organisms/radio-group/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+    >
+      <RadioGroupPrimitive.Item
+        {...rest}
+        id={rest.value}
+        className={clsx(
+          "radio-outer-ring outline-0",
+          "shrink-0 w-[20px] h-[20px] shadow-[0_0_0_1px] shadow-[#D1D5DB] rounded-circle",
+        )}
+      >
+        <RadioGroupPrimitive.Indicator
```
pKorsholm, line 45: this seems to be misaligned (I have made the same comment for Kasper in his pr), just fyi that it should be fixed, i see you are waiting for his pr to be merged first

src/domain/settings/return-reasons/create-reason-modal.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+// the reason props is used for prefilling the form when duplicating
+const CreateReturnReasonModal = ({ handleClose, reason }) => {
+  const { register, handleSubmit, reset } = useForm()
+  const toaster = useToaster()
+  const createRR = useAdminCreateReturnReason()
+
+  useEffect(() => {
+    reset({
+      value: reason.value,
```
pKorsholm, line 17: This causes errors when trying to create a return reason and causes the modal to crash, either this should be `reason?.value` or a value like `""` should be provided. 

I'm not quite sure that I understand why we pass a return reason to the create modal and why we reset every time we update the return reason

src/domain/settings/return-reasons/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+                  setSelectedReason(
+                    findReasonByValue(sorted_return_reasons, value)
+                  )
+                }
+                value={selectedReason?.value}
+              >
+                {sorted_return_reasons.map(reason => (
+                  <RadioGroup.Item
+                    label={reason.label}
+                    sublabel={reason.value}
```
pKorsholm, line 71:
```suggestion
                    sublabel={`(Value: ${reason.value})`}
```

src/domain/settings/return-reasons/detail.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const handleDeletion = async () => {
+    deleteRR.mutate(null)
+  }
+
+  const onSave = data => {
+    if (data.label === "") return
+    updateRR.mutate(data, {
+      onSuccess: () => {
+        toaster("Successfully updated return reason", "success")
+        navigate("/a/settings/return-reasons/")
```
pKorsholm, line 43:
```suggestion
```

This shouldn't be necessary with the caching updates

pKorsholm (commented):
remember to add `"@radix-ui/react-radio-group": "^0.1.4",` to dependencies in `package.json`

src/components/organisms/radio-group/index.tsx
zakariaelas: right! made a tiny improvement by changing the `align-items` property from `baseline` to `flex-start` and used negative margins to kinda "pull up" the div containing the label + description. It might not be pixel perfect aligned but let's investigate that later down the road if that's okay with you :+1: What do you think?

src/domain/settings/return-reasons/create-reason-modal.tsx
zakariaelas: good catch! As I mentioned in my PR, the `CreateReturnReasonModal` is used in two scenarios: 
1. Creating a return reason
2. Duplicating a return reason

For 2. a reason is passed to prefill the input fields, the user can then make the necessary changes and click "create" to create a new return reason. Hope this explains why the component takes a `reason` as prop. I added types for the props and renamed `reason` to `initialReason` to hopefully make this less confusing.

The useEffect would only run once when the `reason` prop is passed down to set the form values, but I think this doesn't communicate the intent in a good way, so I removed that and used `react-hook-form`'s `defaultValues`, which is the idiomatic way of prefilling form fields on initial render. Hope this clears the confusion a bit. 

Let me know what you think and thanks for the comments!

src/domain/settings/return-reasons/detail.tsx
zakariaelas: good catch! thanks!

src/components/organisms/radio-group/index.tsx
kasperkristensen: Just a quick comment on this @zakariaelas, Ive fixed it in my PR by simply changing `items-baseline` to `items-start` on the RadioGroup.Item component. That way we dont need to do any magic negative margins

src/domain/settings/return-reasons/create-reason-modal.tsx
pKorsholm: ah ofc. that makes sense 😄  i must have missed it above, I think thats a very nice way of solving it and communicating the intent

src/components/organisms/radio-group/index.tsx
zakariaelas: alrighty @kasperkristensen, that sounds perfect!

src/domain/settings/return-reasons/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+                    description={reason.description}
+                    className="mt-xsmall"
+                    value={reason.value}
+                  />
+                ))}
+              </RadioGroup>
+            )}
+          </div>
+        </BodyCard>
+        <ReturnReasonDetail reason={selectedReason} />
```
olivermrbl, line 81: nit (suggestion): 
```suggestion
        {selectedReason && <ReturnReasonDetail reason={selectedReason} />}
```

To avoid showing the detail pane in case of no return reasons at all.

olivermrbl (approved):
LGTM, really nice 👍 

A bit of nitpicking and a couple of suggestions.

src/domain/settings/return-reasons/index.tsx
zakariaelas: super good catch!

src/domain/settings/return-reasons/detail.tsx
```diff
@@ -0,0 +1,123 @@
...
+    handleOpen: handleOpenPrompt,
+  } = useModal()
+  const { register, reset, handleSubmit } = useForm()
+  const toaster = useToaster()
+  const deleteRR = useAdminDeleteReturnReason(reason?.id)
+  const updateRR = useAdminUpdateReturnReason(reason?.id)
+
+  const handleDeletion = async () => {
+    deleteRR.mutate(null)
+  }
```
olivermrbl, line 36: nit: Reset `selectedReason` in parent component or refresh data, since right now, if you delete the currently selected reason, it stays in the dom.

olivermrbl (commented):
Sorry, missed a minor one previously @zakariaelas! Let me know if it makes sense 😅

src/domain/settings/return-reasons/detail.tsx
zakariaelas: good catch @olivermrbl! Deletion is triggered in the child component so two ways we can approach this I think:

1. Create a callback function `onDelete` in the parent and pass it down to the `ReturnReasonDetail` component so it can fire it after the deletion takes place
2. Listen to when the return reasons are refetched, then change the selected return reason from the parent. As we've discussed in the past, the refetch happens automatically after a deletion.

Approach 1. is totally valid, but I'm not the biggest fan because in my opinion it adds a little bit of noise if you're reading the `ReturnReasonDetail` component. You'll find an `onDelete()` function being called, and you'll have to go back to see what that function does, then you'll have two components that are changing state (`selectedReason`) instead of one. I personally much rather have whatever `onDelete()` wants to do, to just do it in the parent since it's the one creating and managing the `selectedReason` state without the need to pass a callback down its child.

Took advantage of those react-query `onSuccess` callbacks to do the sorting, as well as figure out which element to select when a query is first fetched + refetched. Just wanted to try it out at first, but then noticed that it reads nicely compared to approach 1 and is much cleaner (no `useEffect`s needed).

Might be overthinking a little here :sweat_smile: but thought I'd share nevertheless.

olivermrbl: Very neat! Def. think this is the better approach @zakariaelas - nicely done! 👍

PR #241, opened 2022-01-16
Dynamic height in useComputedHeight hook

olivermrbl:


kasperkristensen (approved):
LGTM, works like a charm 😄

adrien2p:
Why dont take advantage of https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver

adrien2p:
what will be the use cases for those hooks ?

olivermrbl:
> what will be the use cases for those hooks ?

One use case is for controlling the height of our layout panes. You can read a longer discussion thread on it [here](https://github.com/medusajs/admin/pull/234).

PR #242, opened 2022-01-16
Allow control of BodyCard styling

olivermrbl:
Add `className` prop to `BodyCard` to allow for custom styling.

Example use-case:
If the `<BodyCard/>` inside the two-split pane should not always span full height but fit its content until content expands the view, we can set the height to `h-auto` and `max-h-full`.

### DIFF
--- a/src/components/organisms/body-card.tsx
+++ b/src/components/organisms/body-card.tsx
@@ -1,3 +1,4 @@
+import clsx from "clsx"
 import React from "react"
 import { useScroll } from "../../hooks/use-scroll"
 import Button from "../fundamentals/button"
@@ -11,6 +12,7 @@ type BodyCardProps = {
     onClick: (e: React.MouseEvent<HTMLButtonElement>) => void
   }[]
   actionables?: ActionType[]
+  className?: string
 }
 
 const BodyCard: React.FC<BodyCardProps> = ({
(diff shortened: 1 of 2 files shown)


kasperkristensen (commented):
LGTM with small nit:

I suggest we change it to something along the lines of

```tsx
type BodyCardProps = {
  title?: string
  subtitle?: string
  events?: {
    label: string
    onClick: (e: React.MouseEvent<HTMLButtonElement>) => void
  }[]
  actionables?: ActionType[]
} & React.HTMLAttributes<HTMLDivElement>

const BodyCard: React.FC<BodyCardProps> = ({
  title,
  subtitle,
  events,
  actionables,
  children,
  className,
  ...rest
}) => {
  const { isScrolled, scrollListener } = useScroll({ threshold: 16 })
  return (
    <div
      className={clsx(
        "rounded-rounded border bg-grey-0 border-grey-20 h-full overflow-hidden flex flex-col min-h-[350px] w-full relative",
        className
      )}
      {...rest}
    >
      // ....
```

That way we get autocompletion for Tailwind classes as our IDE will be able to understand that className is an actual className prop. It also allows us to pass other HTML attributes to the component through `...rest` if needed.

pKorsholm (approved):
lgtm 😄

PR #243, opened 2022-01-17
feat: revamp banner card

kasperkristensen:
**What**

- Adds a new BannerCard molecule, with subcomponents Description and Footer.

image

- Also adds a GiftCardBanner organism that extends BannerCard

image

- Included in this is also some new Atoms and molecules: `DenominationBadge />`, `<StatusIndicator />` and `<DenominationGrid />`.
- As part of the `<DenominationGrid />` I've also implemented a new hook `useObserveWidth`, that listens to the width of a `div`. This is used to calculate how many `<DenominationBadge />` components we can display in the "grid" and if there is any remainders we also display a badge with `"+<remainder>"`. @olivermrbl This is not completely inline with what we discussed earlier with Ludvig, but this approach is more responsive in my opinion, but let me know what you think.

image
image

You can view the components in Storybook as they have not been added to any views yet.

**Also in this PR**
- Changes to prices utility to calculate display amounts and persistent amounts using the currencies utility.
- Also updates the currencies utility, by correcting some errors (Updated name of one currency, as it was changed in 2012. And updated decimal points for some currencies, that was incorrect).


src/components/atoms/denomination-badge/index.tsx
```diff
@@ -0,0 +1,26 @@
+import React from "react"
+
+type DenominationBadgeProps = {
```
pKorsholm, line 3: There is a badge component in fundamentals, we should add the variant to that

src/components/atoms/status-indicator/index.tsx
```diff
@@ -0,0 +1,27 @@
+import clsx from "clsx"
+import React from "react"
+
+type StatusIndicatorProps = {
+  ok: boolean
+  okText: string
+  notOkText: string
+}
+
+const StatusIndicator: React.FC<StatusIndicatorProps> = ({
```
pKorsholm, line 10: We should consolidate this with the `fundamentals/status-dot` i like the name `status indicator` a lot better 😄

src/components/molecules/actionables.tsx
```diff
@@ -28,7 +28,7 @@ const Actionables: React.FC<ActionablesProps> = ({ actions }) => {
     <div>
       <DropdownMenu.Root>
         <DropdownMenu.Trigger asChild>
-          <Button variant="ghost" size="medium">
+          <Button variant="ghost" size="small" className="py-0">
             <MoreHorizontalIcon />
```
pKorsholm, line 32: According to the design this should be a square icon 32 x 32 with horizontal dots of size 20. I suggest we do something like this, but if you fancy we can go with fixed widths as well ie. `w-[32px]`
```suggestion
          <Button 
            variant="ghost" 
            size="small" 
            className="w-8 h-8 p-1.5 focus:border-none focus:shadow-none"
          >
            <MoreHorizontalIcon size={20} />
```
```

File: src/components/molecules/denomination-grid/index.tsx [hunk 4]
```diff
@@ -0,0 +1,83 @@
...
+  >([])
+  const [remainder, setRemainder] = useState(0)
+
+  useEffect(() => {
+    if (defaultDenominations.length) {
+      setDenominations(defaultDenominations.slice(0, columns))
+      setRemainder(defaultDenominations.length - columns)
+    }
+
+    return
```
pKorsholm, line 57: rm
```suggestion
    
```

pKorsholm (commented):
Looks very nice! just a few comments

src/components/molecules/denomination-grid/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+  const width = useObserveWidth(containerRef)
+
+  const [columns, setColumns] = useState(7)
+
+  // We estimate that a Badge is rouglhy 70px and use this to calulate how many we can fit in a row
+  // It is a bit of a "magic" number, but it works for now.
+  useEffect(() => {
+    if (width) {
+      setColumns(Math.floor(width / 70) - 1)
```
zakariaelas, line 31: In the case where width < 70, columns will be set to -1, which will make the denominations behave in a funny way. You can try it by resizing the viewport to a smaller size (around < 330px)

One nifty solution would be to do this instead: `setColumns(Math.max(Math.floor(width/70) - 1, 1)`. This would basically say: "always show at least 1 denomination".

src/components/molecules/denomination-grid/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const [denominations, setDenominations] = useState<
+    { amount: number; currencyCode: string }[] | null
+  >([])
+  const [remainder, setRemainder] = useState(0)
+
+  useEffect(() => {
+    if (defaultDenominations.length) {
+      setDenominations(defaultDenominations.slice(0, columns))
+      setRemainder(defaultDenominations.length - columns)
+    }
```
zakariaelas, line 55: nit: I believe the states here (+columns) could be derived because everything other piece of state is derived from the `width` we get from the `useObserveWidth`. Meaning this would be enough, no `useEffect`s needed:
```jsx
const columns = Math.max(Math.floor(width / 70) - 1, 1)
const denominations = defaultDenominations.slice(0, columns)
const remainder = defaultDenominations.length - columns
```

src/hooks/use-observe-width.ts
```diff
@@ -0,0 +1,21 @@
...
+  )
+
+  useEffect(() => {
+    observer.current.observe(ref.current)
+
+    return () => observer.current.unobserve(ref.current)
+  }, [ref, observer])
+
+  return currentWidth
+}
```
zakariaelas, line 21: very nice!

src/components/molecules/banner-card/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+    onClick: (e: React.MouseEvent<HTMLButtonElement>) => void
+  }
+}
+
+const BannerCard: React.FC<BannerCardProps> & {
+  Description: React.FC<BannerCardDescriptionProps>
+  Footer: React.FC
+} = ({ title, thumbnail, actions, children }) => {
+  return (
+    <div className="rounded-rounded border bg-grey-0 border-grey-20 w-full p-xlarge">
```
zakariaelas, line 22: I suggest we have different paddings for different screen sizes, for example: `medium:p-xlarge xsmall:p-base`. To check with Ludvig of course, but I think we can eat up some of the paddings when we're on small screens

src/components/molecules/denomination-grid/index.tsx
zakariaelas: even smaller nit: additionally, you could extract everything related to the denominations in a `useDenominationClamp` and just use it in the `DenominationGrid`. To what extent you'd want it to be reusable is up to you, but simply extracting into a hook without much re-usability would be good enough to clean up the `DenominationGrid` I think

src/components/molecules/denomination-grid/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const defaultDenominations = useMemo(() => {
+    return (
+      variants.map(variant => {
+        const price = variant.prices.find(
+          price => price.currency_code === defaultCurrency
+        )
+        return { amount: price?.amount, currencyCode: defaultCurrency }
+      }) ?? []
+    )
+  }, [variants, defaultCurrency])
```
zakariaelas, line 44: nit: what do you think about moving this up to the `GiftCardBanner` component and making the `DenominationGrid` only take a prop called `denominations` (or any better name) of type `Array<{ amount: number, currencyCode: string }>`? I feel like `DenominationGrid` shouldn't have to know about how to extract a price from a variant. The `useMemo` might also be an indicator that it might fit better somewhere else. Let me know what you think :smile:

zakariaelas (commented):
Great work Kasper! Super clean and I love the `useObservableWidth` :sunglasses:. Left a few comments (mostly nits) but LGTM :+1:

src/components/molecules/banner-card/index.tsx
kasperkristensen: Agree, the component will still need some additional work once we implement a truly responsinve layout, as it is a bit cramped on mobile screens even with adjustments to padding. But I think this is a good way to make it less cramped on a smaller computer screen/browser window :smile

src/components/molecules/actionables.tsx
kasperkristensen: definitely not inline with the design, but the issue is that the design of the button (if it is set to 32x32) makes it so large that it pushes everything else out of place. I could do something like setting the positioning to absolute so it does not affect everything around it, but in that case we would have to do some weird magic numbers to make sure that it sits in the proper position. The issue can be seen on this image, as this does not translate that well into css:

image

But changing the entire look of Actionables was definitely not a good idea 😅 I've made a change that brings it closer in line with the design without changing the look of Actionables, and then I will check in with Ludvig, but I think this should work just fine for now 😄

src/components/molecules/denomination-grid/index.tsx
kasperkristensen: definitely, I have moved it to GiftCardBanner and made it return a string array that we pass along to the now renamed TagGrid

src/components/molecules/banner-card/index.tsx
kasperkristensen: Agree, the component will still need some additional work once we implement a truly responsinve layout, as it is a bit cramped on mobile screens even with adjustments to padding. But I think this is a good way to make it less cramped on a smaller computer screen/browser window 😄

olivermrbl (approved):
LGTM, super nice work!

Just a minor addition 👍

PR #244, opened 2022-01-17
Feat: Revamp gift card list view

pKorsholm:
**What** 
- Add list pane for gift card overview
- for now include filtering and sorting fields for example (will remove before merging)
- add cell for sorting with icons
  - here we could have a discussion about what is the right approach, I see two ways outlined below
- stories for list pane

image

### sorting cell implementation discussion
I see there are two ways which we can manage state when implement a sorting cell. In one way we let the cell itself determine the state (i.e. which function to call next as well as arrow up/arrow down). I think this inevitably will lead to errors in the UI because the state cannot be reset from outside the component.

In the other way we provide the component with a string/enum being either 'ASC' for ascending, 'DESC' for descending or undefined for not sorted. Furthermore we would provide the component with a method of setting this string in the parent component, meaning we would pass state and setState for each of the headCells into the table. We would then be able to invoke the correct search by utilizing `useEffect` listening on the sorting states of each of the cells, and the parent component would be responsible for invoking the api to refetch records in sorted order. 

Lastly we could choose a hybrid approach between the two using state as described in the previous paragraph, but also utilizing a single method like `onSortClicked` which would allow the sorting cell to invoke a refetch when it's clicked. The parent component would then know the state of the sorting cells and be able to construct a query accordingly. Furthermore this, unlike the solution described in the previous paragraph would allow us to control the state independently of refetching from the API.

I have implemented the last option as I think it is the most flexible and enables the most versatile use of the listpane component in future implementations.


zakariaelas (commented):
Great work Philip! Left a few tiny comments.

Also, tables are a bit tricky to deal with from my experience especially when it gets to server-side filtering, sorting, search, etc. In my opinion, this could be a great candidate for [react-table](https://react-table.tanstack.com/), which is a headless library for managing tables in react. It's battle-tested, provides a lot of functionality, and solves a lot of the common problems faced when working with tables in a nice way. I think building on top of it will essentially allow us to have a reliable solution we can reuse across different screens which will be super cool! I believe you can also reuse some of your work on filtering so it's not a full replacement for your previous work. I think now might be a good time before getting into products and orders (probably the most complicated tables we'll have on Admin). 

Let me know what you think, perhaps we're already too deep in with the current implementation, but let's get @kasperkristensen, @olivermrbl, and @srindom also involved in the discussion :smile:

olivermrbl (approved):
LGTM, great work Philip!

I would argue, that we should merge this in now, and not worry about the `react-table` implementation for now, since it's still WIP. So let's see if we have time next week to switch out this table with a battle-tested `react-table` implementation.

PR #248, opened 2022-01-18
feat: revamp textarea

zakariaelas:
### What
- adds textarea 
- adds a story for textarea
- fixes typescript errors + prettier for `InputContainer` and `input.stories.tsx`

### How
- used @pKorsholm's work on `Input` as reference
- used `@tailwindcss/line-clamp` (thanks @kasperkristensen!) and extended it with a custom class to allow for dynamic values in `line-clamp` based on textarea's `rows` value

_Note: clamping the multiline textarea works by using the [line-clamp css property](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-line-clamp). I noticed some issues with it when using Firefox, but as far as I have tried, it works well with Chrome and Brave. So please if you have Safari/Firefox, try to run it there and see if the clamping behaves as expected._

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "@radix-ui/react-radio-group": "^0.1.4",
     "@rebass/forms": "^4.0.6",
     "@svgr/webpack": "^5.4.0",
+    "@tailwindcss/line-clamp": "^0.3.1",
     "axios": "^0.24.0",
     "clsx": "^1.1.1",
     "copy-to-clipboard": "^3.3.1",
--- a/src/assets/styles/global.css
+++ b/src/assets/styles/global.css
@@ -305,15 +305,6 @@
   .vice-city {
     @apply bg-gradient-to-tr from-vice-start to-vice-stop;
   }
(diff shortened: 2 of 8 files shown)


pKorsholm (commented):
Clamping seems to be a pretty universal problem that is posted to stack overflow almost daily, though there doesn't seem to be a good solution. I dont see the ellipsis in firefox and safari.

Other than that it looks great 😅

olivermrbl:
> Clamping seems to be a pretty universal problem that is posted to stack overflow almost daily, though there doesn't seem to be a good solution. I dont see the ellipsis in firefox and safari.
> 
> Other than that it looks great 😅

Yeah, everyone seems to be struggling with this. I actually think this is fine for now even though the ellipsis is missing. We need this component for a lot of other things, so I'd say we get this is now and take it up at a later point 😅 

Is that okay with you guys? @zakariaelas @pKorsholm

zakariaelas:
@olivermrbl  sounds perfect!

PR #253, opened 2022-01-19
add: tooltip + stories + adapt input

zakariaelas:
### What

- adds a generic tooltip component
- adds an info tooltip component (a tooltip which has an info icon as a trigger)
- adapts the `Input` component to use an info tooltip next to its label
- adds info + alert icons and their stories
- adds stories for tooltips
- adds a story for info tooltip
- adds a story for an input with tooltip
- adds a story for an input with custom tooltip

### How 
-  uses Radix's tooltip component

### DIFF
--- a/.storybook/preview-head.html
+++ b/.storybook/preview-head.html
@@ -1,0 +1,8 @@
+<style>
+  html, body {
+    height: 100%;
+  }
+  #root {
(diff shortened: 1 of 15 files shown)


olivermrbl (approved):
LGTM, great work Zakaria! 👍

PR #254, opened 2022-01-19
Feat: revamp discount list

pKorsholm:
**What**
- revamp of `domain/discount/index.tsx`

**How**
- add `BodyCard` with an actionable which redirects to `/a/discounts/new`
- add discount-table template which does all fetching og discounts and setting of url
  - query parameters for `list-discount` call is created based on the url
  - filtering options based on `type` creates a new `GET` request based on the type and fetches the discounts, sets the url and updates the list view
  - search capabilities which utilize the `list-discounts` api endpoint
  - each record links to discount details 
  - each record has the following actionables:
    - *edit discount* which redirects to the `/a/discounts/<id>`, same as the row
    - *duplicate discount* which duplicates the record with a new code begin: `old code DUPLICATE`
    - *delete discount* which shows a delete prompt and removes the discount if the user accepts

image

*note* the type filtering is dependent on pr [#1002](https://github.com/medusajs/medusa/pull/1002) in the core project


src/components/templates/discount-table.tsx
```diff
@@ -0,0 +1,368 @@
...
+    isLoading,
+    count,
+    toaster,
+    ...rest
+  } = useMedusa("discounts", {
+    search: {
+      is_dynamic: "false",
+      ...filtersOnLoad,
+    },
+  })
```
olivermrbl, line 88: Should use `useAdminDiscounts` hook

src/components/templates/discount-table.tsx
```diff
@@ -0,0 +1,368 @@
...
+
+  //   refresh({
+  //     search: {
+  //       ...queryParts,
+  //       is_dynamic: showDynamic,
+  //     },
+  //   }).then(() => {
+  //     setOffset(updatedOffset)
+  //   })
+  // }
```
olivermrbl, line 154: rm

src/components/templates/discount-table.tsx
```diff
@@ -0,0 +1,368 @@
...
+    Medusa.discounts
+      .create(newDiscount)
+      .then(() => {
+        toaster("Successfully created discount", "success")
+        handleCheckbox()
+      })
+      .catch((error) => {
+        toaster(getErrorMessage(error), "error")
+      })
+      .finally(() => setLoading(false))
```
olivermrbl, line 187: Should use `useAdminCreateDiscount` hooks

src/components/templates/discount-table.tsx
```diff
@@ -0,0 +1,368 @@
...
+          onClick: () => {},
+        },
+        {
+          title: "Scheduled",
+          count: 0,
+          onClick: () => {},
+        },
+      ],
+    },
+  ]
```
olivermrbl, line 246: Should we delete these for now? I could imagine that these would be refactored when we start migrating to `react-table`

src/components/templates/discount-table.tsx
```diff
@@ -0,0 +1,368 @@
...
+          onDelete={() =>
+            Medusa.discounts.delete(deleteDiscount.id).then(() => {
+              searchQuery()
+            })
+          }
+          handleClose={() => setDeleteDiscount(undefined)}
+        />
+      )}
+
+      <div className="flex w-full justify-between inter-small-regular text-grey-50">
```
olivermrbl, line 356: Not quite in alignment with the design. Think it needs `mt-8`:
image

src/domain/discounts/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+  const { discounts, refresh, isReloading, isLoading } = useMedusa(
+    "discounts",
+    {
+      search: {
+        is_disabled: "false",
+        is_dynamic: "false",
+        ...filtersOnLoad,
+      },
+    }
+  )
```
olivermrbl, line 37: Should use `useAdminDiscounts` hook

olivermrbl: Do we need this? As far as I see, we are doing the fetching in the actual discount table component. If not, we should remove this and all of the below querying and filtering :)

olivermrbl (changes requested):
Really great job Philip! Similar to the gift cards list overview, let's hold off with the `react-table` implementation and see if time allows for it next week. 👍

olivermrbl (approved):

PR #259, opened 2022-01-22
feat: DetailsCollapsible

olivermrbl:
**What**
Adds a collapsible organism for showing/hiding details.

**How**
Use `@radix-ui/collapsible`

Decided to add this as an organism, since it's already being used multiple times in the design. Considered adding a more generic collapsible molecule, but Radix UI's collapsible is already quite customizable, so did not find this relevant. Alternatively, we could also completely remove this component and use Radix directly. Let me know, what you think :)

### DIFF
--- a/src/components/organisms/details-collapsible/index.tsx
+++ b/src/components/organisms/details-collapsible/index.tsx
@@ -4,21 +4,41 @@ import React, { useState } from "react"
 import ArrowDownIcon from "../../fundamentals/icons/arrow-down-icon"
 import ArrowUpIcon from "../../fundamentals/icons/arrow-up-icon"
 
-const DetailsCollapsible = ({ children }) => {
+type DetailsCollapsibleProps = {
+  rootProps?: RadixCollapsible.CollapsibleProps
+  triggerProps?: RadixCollapsible.CollapsibleTriggerProps
+  contentProps?: RadixCollapsible.CollapsibleContentProps
+  children: React.ReactNode
+}
+
+const DetailsCollapsible = ({
+  rootProps,
+  triggerProps,
+  contentProps,
+  children,
+}: DetailsCollapsibleProps) => {
   const [open, setOpen] = useState(false)
 
   const Icon = open ? ArrowUpIcon : ArrowDownIcon
   const label = open ? "Hide additional details" : "Show additional details"
 
   return (
-    <RadixCollapsible.Root onOpenChange={(state) => setOpen(state)}>
-      <RadixCollapsible.Trigger className={clsx("ml-4", { ["mb-6"]: open })}>
+    <RadixCollapsible.Root
+      {...rootProps}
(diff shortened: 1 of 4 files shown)


src/components/organisms/details-collapsible/index.tsx
```diff
@@ -0,0 +1,26 @@
+import * as RadixCollapsible from "@radix-ui/react-collapsible"
```
pKorsholm, line 1: Should we streamline our radix-ui imports? i.e. always use something like `Radix[componentName]` for clarity?

pKorsholm (approved):
LGTM, very nice

src/components/organisms/details-collapsible/index.tsx
olivermrbl: I like this idea! Don't think we should prioritize it for the already merged features, but let's see if we can make it happen for future PRs 👍

zakariaelas:
LGTM :+1: Super cool!

zakariaelas (approved):
LGTM +1 Super cool!

src/components/organisms/details-collapsible/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+const DetailsCollapsible = ({ children }) => {
+  const [open, setOpen] = useState(false)
+
+  const Icon = open ? ArrowUpIcon : ArrowDownIcon
+  const label = open ? "Hide additional details" : "Show additional details"
+
+  return (
+    <RadixCollapsible.Root onOpenChange={(state) => setOpen(state)}>
+      <RadixCollapsible.Trigger className={clsx("ml-4", { ["mb-6"]: open })}>
```
pKorsholm, line 15: We should probably remove the `ml-4` from the trigger and potentially add it through a `triggerClassname` property instead to give more flexibility in terms of styling

(i.e. in the manage gift card pr the margin should be ml-2 to match the design)

src/components/organisms/details-collapsible/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+import React, { useState } from "react"
+import ArrowDownIcon from "../../fundamentals/icons/arrow-down-icon"
+import ArrowUpIcon from "../../fundamentals/icons/arrow-up-icon"
+
+type DetailsCollapsibleProps = {
+  rootProps?: RadixCollapsible.CollapsibleProps
+  triggerProps?: RadixCollapsible.CollapsibleTriggerProps
+  contentProps?: RadixCollapsible.CollapsibleContentProps
+  children: React.ReactNode
+}
```
olivermrbl, line 12: Let me know, what you think of this @pKorsholm 😄

pKorsholm: I think it's good 😄  just so we have some control from outside the component as well

pKorsholm (approved):
lgtm

PR #261, opened 2022-01-24
feat: add denominations modal

zakariaelas:
### What

- adds edit denominations modal
- changes the modal close from a `<span>` to a `<button>`
- uses `CurrencyInput` component 

### Files to review
- `src/components/organisms/edit-denominations-modal/*`


src/components/organisms/edit-denominations-modal/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+    return () => {
+      const newDenominations = denominations.slice()
+      newDenominations.splice(index, 1)
+      setDenominations(newDenominations)
+    }
+  }
+
+  const appendDenomination = (newDenomination) => {
+    setDenominations([...denominations, augmentWithId(newDenomination)])
+  }
```
olivermrbl, line 72: nit: Think it would be cool to add two rules to this operation:
1. When appending a new denomination use a currency code that is not already present among the other denominations
2. Add a limit on number of denominations that correlates to the number of different currencies in store

image

olivermrbl: E.g. if I have DKK, USD, and EUR as store currencies, the above image should disable "+ Add a price" when there are three rows and row 3 should be EUR instead of USD

src/components/organisms/edit-denominations-modal/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+type EditDenominationsModalProps = {
+  defaultDenominations: PriceType[]
+  handleClose: () => void
+  onSubmit: (denominations: PriceType[]) => void
+  defaultNewAmount?: number
+  defaultNewCurrencyCode?: string
+  currencyCodes?: string[]
+}
+
+const augmentWithId = (obj) => ({ ...obj, id: obj.id ? obj.id : uuidv4() })
```
olivermrbl, line 25: This tampers a bit with the data when we send the denominations to the backend because we are providing the id ourselves instead of letting Postgres generate it.

We could say, that the parent component is required to normalize the data from the denominations modal to fit the backend schema, but that would not be very obvious to developers coming from the outside using the component. Maybe we could add a different prop `indexId` that is dedicated to solving the rendering issue and then remove that prop on all denominations when submitting the data? 

Let me know what you think 👍

olivermrbl (commented):
Great work @zakariaelas! Couple of comments 👍

src/components/organisms/edit-denominations-modal/index.tsx
zakariaelas: great idea oli :+1: !

src/components/organisms/edit-denominations-modal/index.tsx
zakariaelas: @olivermrbl I pushed a fix for the above. I decided not to interact with our store explicitly from the modal and instead, act entirely on the `currencyCodes` prop. This allows the EditDenominationsModal to stay as "dumb" as possible. You could achieve what you mentioned above by doing something like this from any component that consumes the modal: 
```jsx
const Component = () => {
  const { store } = useAdminStore()
  
  return (
    <EditDenominationsModal 
        currencyCodes={store.currencies.map(c => c.code)} 
        onSubmit={console.log} 
        handleClose={handleClose}
    />
  )
}
```

What do you think about this approach? Do you prefer keeping the modal as "pure" as possible, or would you rather we call `useAdminStore()` from the modal? 

Also, feel free to use the storybook link deploy link to play around with this :smile:

olivermrbl: I really like this @zakariaelas! Agree, that we should strive to keep modals pure - great work 👍

olivermrbl (approved):
LGTM! 😍

PR #262, opened 2022-01-24
feat: revamp currency input

kasperkristensen:
**What**

- Adds new `CurrencyInput` component
  - Can be used as either a dropdown select, or as a ReadOnly div if the currency should not be changeable. This is controlled using the `readOnly` prop.

- Has a subcomponent of `CurrencyInput.AmountInput`
  - Uses context from CurrencyInput component to check how to calculate display and persistent amounts (fx if currency is KRW then the divider/multiplier should be 1, if the currency is USD then it should be 100). Therefore, it is no longer necessary to multiply and divide the amount as previously. As an example you can now pass `variants[0].prices[0].amount` to this component and you will not have to do `variants[0].prices[0].amount / 100` to get the proper display amount, or `amount * 100` to get the correct persistent amount.

**Open for suggestions**
- Currently you will need to write handler methods to handle both the currency and amount changing, in order for it to work with `useForm`. Such as:

```tsx
const handleAmountChange = (fieldName: string, amount?: number) => {
    setValue(fieldName, amount)
}

// ...

<div className="grid grid-cols-1 medium:grid-cols-2 gap-base">
                  <CurrencyInput
                    currentCurrency={region.currency_code}
                    readOnly
                  >
                    <CurrencyInput.AmountInput
                      label="Price"
                      onChange={(v) =>
                        handleAmountChange(
                          "requirements.min_subtotal.amount",
                          v
                        )
                      }
                      amount={undefined}
                    />
                  </CurrencyInput>
                  <CurrencyInput
                    currentCurrency={region.currency_code}
                    readOnly
                  >
                    <CurrencyInput.AmountInput
                      label="Price"
                      onChange={(v) =>
                        handleAmountChange(
                          "requirements.max_subtotal.amount",
                          v
                        )
                      }
                      amount={undefined}
                    />
                  </CurrencyInput>
                </div>
```
I have not been able to figure out a way where we can just pass `{...register( "requirements.max_subtotal.amount")}`. This would definitely be preferable. So if you have any ideas how we could do it please shoot, as this would also be great to implement for the `Select` component.

**Note**
- Merge after #243


src/domain/settings/regions/details.tsx
```diff
@@ -255,14 +246,20 @@ const RegionDetails = ({ id, onDelete, handleSelect }) => {
...
                 />
-                <Select
+                {/* <Select
                   enableSearch
                   label="Currency"
                   name="currency_code"
                   options={currencies}
                   value={selectedCurrency}
                   onChange={handleChangeCurrency}
                   className="mb-base"
```
olivermrbl, line 256: rm

olivermrbl (approved):
LGTM, really, really great work @kasperkristensen ! 👍

src/components/organisms/currency-input/index.tsx
```diff
@@ -0,0 +1,224 @@
...
+  step = 1,
+  allowNegative = false,
+  onChange,
+}) => {
+  const { currencyInfo } = useContext(CurrencyContext)
+  const [value, setValue] = useState<string | undefined>(
+    amount ? `${amount}` : undefined
+  )
+  const inputRef = useRef<HTMLInputElement | null>(null)
+
```
pKorsholm, line 142:
```suggestion

  const [loading, setLoading] = useState(true)
  useEffect(() => setLoading(false), [])
  
```

src/components/organisms/currency-input/index.tsx
```diff
@@ -0,0 +1,224 @@
...
+    if (currencyInfo && amount) {
+      setValue(`${displayAmount(currencyInfo.code, amount)}`)
+    }
+  }, [amount, currencyInfo])
+
+  /**
+   * Returns the persited amount for the current currency
+   */
+  useEffect(() => {
+    let persistedAmount: number | undefined = undefined
```
pKorsholm, line 160:
```suggestion
    if (loading) {
      return
    }
    
    let persistedAmount: number | undefined = undefined
```

pKorsholm (approved):
Very nice
We should add a blocking variable to ensure rendering doesn't cause invocation of `onChange`

pKorsholm:
Also it seems there is a float rounding error in the code, i.e. rounding down because the value can get set to .999999999 and not the whole number, which in turn rounds down and when is incremented back to .999999 next time "+" is clicked.

PR #263, opened 2022-01-24
Feat: Revamp filter dropdown

pKorsholm:
**What** 
- revamp of filter dropdown
- revamp of filter dropdown items
- revamp of filter dropdown date pickers

**How**
- The trigger is passed as a parameter to the container
- The children of the container is shown in the list

**Files to review** 
`src/components/molecules/filter-dropdown/container.tsx`
`src/components/molecules/filter-dropdown/item.tsx`
`src/components/molecules/filter-dropdown/save-field.tsx`
`src/components/molecules/tag-input/index.tsx`
`src/domain/products/filter-dropdown.tsx`

image

image

dependent on #257


kasperkristensen (changes requested):
Looks very good! Not sure if it is due to the the new filter menu being used together with the old one, but I am seeing some strange behaviour fx when I choose to filter by date then it first show a date picker above the menu, and after I select a date a new date picker pops up on the left side of the screen. And the second date picker is only able to be cancelled.

image

Also might be nice to create a Story for this as the component is not used in full anywhere (at least not as far as I can see) so it's a bit difficult to review if the styling is correct 😄

pKorsholm:
The strange behavior is due to the old menu, if you copy the date filter menu item into product filters its not a problem

src/components/molecules/filter-dropdown/item.tsx
```diff
@@ -0,0 +1,423 @@
...
+  options,
+  open,
+  setFilter,
+  existingDate,
+  existingFilter,
+  filterTitle,
+}) => {
+  const [currentFilter, setCurrentFilter] = useState(
+    existingFilter || undefined
+  )
```
olivermrbl, line 180: It seems, that the date filter does not hold the value of the applied filter. So when you have applied a filter and re-open the date filter item, the value of the applied filter is no longer present. 

image

olivermrbl: @pKorsholm Do you have an idea, why this happens?

olivermrbl (approved):
LGTM, great work! The missing persistency of the date filter will be added as part of #286 🚀

PR #265, opened 2022-01-25
feat: Customer list overview

olivermrbl:
**What**
Adds Customer list overview
Adds `TablePagination` component in table

**How**
`react-table` PoC  implementation


src/domain/customers/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+          </div>
+        </BodyCard>
+      </div>
+    </div>
+  )
+}
+
+const Customers = () => {
+  return (
+    <Router>
```
pKorsholm, line 32:
```suggestion
    <Router className="h-full" >
```

pKorsholm (approved):
This looks great 😄  I have only a few comments i've left as suggestions

src/domain/customers/index.tsx
olivermrbl: Adding this class results in the following table view. I ran it by Ludvig and he would rather have the "shrinked" table so he's gonna update the table designs accordingly 👍 
image

pKorsholm: cool 👍  well then just ignore the comment 😄

PR #268, opened 2022-01-25
Feat: Revamp product list

pKorsholm:
**What** 
- product list
- product index

files to review: 
- `src/components/molecules/table/index.tsx`
- `src/components/templates/product-table/index.tsx`
- `src/domain/products/filter-dropdown.tsx`
- `src/domain/products/index.tsx`

*dependent on #263 and #257*

image


src/components/templates/product-table/index.tsx
```diff
@@ -0,0 +1,543 @@
...
+            .then(() =>
+              toaster(
+                `Successfully ${
+                  product.status === "published" ? "unpublished" : "published"
+                } product`,
+                "success"
+              )
+            )
+            .then(() => refetch())
+            .catch((err) => toaster(getErrorMessage(err), "error")),
```
olivermrbl, line 431: Should use `useAdminUpdateProduct` hook

src/components/templates/product-table/index.tsx
```diff
@@ -0,0 +1,543 @@
...
+        : null
+
+    setTagsFilter(tagsFilter)
+  }
+
+  const filtersOnLoad = qs.parse(
+    window.location.search.charAt(0) === "?"
+      ? window.location.search.substring(1)
+      : window.location.search
+  )
```
olivermrbl, line 145: Filters on load work really well, but it would be great to add the filters being applied on load to the custom filter count. 

So e.g. if I send you the following URL with the collection filter (shown in image below), it will show as "Custom filter: **1**", when another person opens it.

image

src/domain/products/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+          <ProductTable />
+        </BodyCard>
+      </div>
+    </div>
+  )
+}
+
+const Products = () => {
+  return (
+    <Router className="h-full">
```
olivermrbl, line 89:
```suggestion
    <Router>
```

olivermrbl (changes requested):
Really great work and so nice with `react-table`! It's all coming together 🚀

src/components/templates/product-table/index.tsx
```diff
@@ -0,0 +1,543 @@
...
+
+  const columns = useMemo(
+    () => [
+      {
+        Header: "Name",
+        accessor: "title",
+        Cell: ({ row: { original } }) => {
+          return (
+            <div className="flex items-center">
+              <div className="w-[38px] h-[38px] mr-4">
```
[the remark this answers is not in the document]
olivermrbl: Actually, scratch that @pKorsholm.

Just played around with it on customer orders table and adding the following gets us really close to the design. Let me know what you think: 

```
<div className="h-[35px] w-[25px] flex items-center "> // <- The wrapper
   <img className="rounded object-cover">
```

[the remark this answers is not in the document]
pKorsholm: I think this is the solution in terms of hitting the design 

```javascript
 <div className="h-[40px] w-[30px] my-1.5 flex items-center mr-4">
     <img className="h-full object-cover rounded-soft"
```

src/components/templates/product-table/index.tsx
```diff
@@ -0,0 +1,614 @@
...
+            filteringOptions={
+              <ProductsFilter
+                setStatusFilter={setStatusFilter}
+                statusFilter={statusFilter}
+                setCollectionFilter={setCollectionFilter}
+                collectionFilter={collectionFilter}
+                collections={collectionsList}
+                setTagsFilter={toggleFilterTags}
+                submitFilters={submitFilters}
+                tagsFilter={tagsFilter}
```
zakariaelas, line 542: Do you think we can do something to avoid all this? I'm thinking `useReducer` might help here .. you would then pass on your dispatch function + filters object to the `ProductsFilter`. I'm thinking the state/filters could be something like this:
```javascript
{
  "status": { open: false,  value: "" },
  "collections": { open: false, value: "" },
  ...
}
```

Just an initial thought, main idea here is to do something to make the `ProductTable` a bit simpler and shorter

src/components/templates/product-table/index.tsx
```diff
@@ -0,0 +1,614 @@
...
+      ) : (
+        <>
+          <Table
+            filteringOptions={
+              <ProductsFilter
+                setStatusFilter={setStatusFilter}
+                statusFilter={statusFilter}
+                setCollectionFilter={setCollectionFilter}
+                collectionFilter={collectionFilter}
+                collections={collectionsList}
```
zakariaelas, line 539: what do you think about creating a `CollectionsFilter` component that you can use inside the `ProductsFilter` that would be responsible for fetching the collections and rendering them? It could take a prop like `onSelect` or `onFilter` which would update the filters state with the chosen collection(s)

Idea here is to have the `useAdminCollections()` hook isolated in the `CollectionsFilter` instead of having it at the top level.

zakariaelas (commented):
Great work Philip :astonished: :heart_eyes: ! Definitely a meaty ticket :laughing: . Left a few comments mostly about reducing the size of the `ProductTable` component. I think it'd be really great to reduce it a bit in terms of size and responsibility. Let's see if we can extract some functionality into hooks or other components so it becomes a bit easier to extend and maintain in the future :rocket:. So would love to hear your thoughts and suggestions on some of the comments :smile:

zakariaelas (approved):
Love the improvements @pKorsholm! LGTM :+1: w. small comment

olivermrbl (commented):
LGTM, just need to hold off with the merge until we have support in core.

olivermrbl (approved):

PR #272, opened 2022-01-27
feat: revamp discount details and new

kasperkristensen:
**What**

- Revamps discount details and new domains
- Implements two new templates for these view
- Adds StatusSelector component

**Note**

There are a bunch of rules related to discounts, what can be updated which combinations are valid and so on. I think I got them all, but if not then please shout! 😄


src/components/templates/discount-general/index.tsx
```diff
@@ -0,0 +1,209 @@
...
+                    disabled={regionIsDisabled}
+                  />
+                  <InputField
+                    label="Amount"
+                    required
+                    type="number"
+                    placeholder="10"
+                    prefix={discountType === "percentage" ? "%" : undefined}
+                    name="rule.value"
+                    ref={register({ required: true })}
```
olivermrbl, line 138: Should not be required if discount is free shipping, since I think this prevents us from saving the discount. At least that's the behaviour, I am encountering atm

src/components/templates/discount-settings/index.tsx
```diff
@@ -0,0 +1,276 @@
...
+  description: string
+  tooltip?: string
+}
+
+const Section: React.FC<SectionProps> = ({
+  title,
+  description,
+  tooltip,
+  children,
+}) => {
```
olivermrbl, line 259: Would it make sense to refactor this is a molecule?

olivermrbl (commented):
Really great work @kasperkristensen, this was a big ticket 👍 

Added a couple of comments for starters

src/components/templates/discount-general/index.tsx
kasperkristensen: @olivermrbl  Actually think this is due to an error in the validator used in the API that requires that amount is a number always 😅 

```ts
@IsNumber()
  value: number
```

I fixed it for now by setting the value to 0 if free shipping is selected :smile

src/components/templates/discount-settings/index.tsx
kasperkristensen: sure might be useful elsewhere 😄

olivermrbl (approved):
Works really well! Great work Kasper 👍

srindom:
Spotted a few things:

**We should change "Draft" to "Disabled"**

**When creating a discount** 
- select template discount
- edit availability duration
- deselect template discount
- submit still sends `valid_duration`

----
the below are more nice to haves

**When creating a discount**
- select 1 region
- select fixed amount
- would be nice if the amount input changes to currency formatted input

**When creating/editing a discount**
- would be nice if the date/time picker displayed the local time but transformed to UTC when submitting

PR #275, opened 2022-01-28
fix: custom actionables

kasperkristensen:
**What**

- Adds the ability to force a dropdown menu on Actionables. This is controlled by setting `forceDots` on Actionables (defaults to false). This can also be set on BodyCard by setting `forceDropdown`
- Allows to provide a customTrigger on Actionables (and BodyCard), controlled by setting `customTrigger` to a ReactNode such as `<button onClick={() => console.log("custom")}>Custom Button</button>`. This is controlled on BodyCards by setting `customActionable`.

Skærmbillede 2022-01-28 kl  08 54 59
Skærmbillede 2022-01-28 kl  08 55 36

### DIFF
--- a/src/components/molecules/actionables.tsx
+++ b/src/components/molecules/actionables.tsx
@@ -14,7 +14,7 @@ export type ActionType = {
 type ActionablesProps = {
   actions?: ActionType[]
   customTrigger?: React.ReactNode
-  forceDots?: boolean
+  forceDropdown?: boolean
 }
 
 /**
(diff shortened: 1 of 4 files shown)


src/components/molecules/actionables.tsx
```diff
@@ -13,28 +13,38 @@ export type ActionType = {
 
 type ActionablesProps = {
   actions?: ActionType[]
+  customTrigger?: React.ReactNode
+  forceDots?: boolean
```
zakariaelas, line 17: tiny nit: maybe `forceDropdown` instead of `forceDots`

zakariaelas (approved):
LGTM :+1: This will definitely be useful :smile:

PR #277, opened 2022-01-30
feat: Order details

olivermrbl:
**Adds order details**

_Depends on #279 _

**Things not included or missing:**
- Timeline
- Refund modal
- Manual payment modal
- All RMA modals

**Files to review:**
- `src/domain/orders/details/index.tsx`
- `src/domain/orders/details/address-modal.tsx`

<div class="flex">

</div>

PR #281, opened 2022-01-31
Feat/revamp timeline

kasperkristensen:
**What**

- Adds revamped timeline
  - Adds all new types of events
  - Gives a more accurate history, as canceled events are show as both when they were created and canceled.
- Adds CopyToClipboard atom
- Adds EmojiPicker ✅

image
image
image

**Note**
- Most actions on events are not implemented, this will be revisited once RMA modals are in

Merge #277 first


olivermrbl (commented):
This is amazing work @kasperkristensen 😍 Looks really beautiful.

**Suggestion for the review process:**
Since this is a huge but isolated PR, I would suggest that we go about it like we've done with some of the PRs for the core (read Shopify plugin, hooks). 

So we do a rough review to find (if any) bigger bugs, improvements, and fixes, but otherwise let the community and others find them for us as we go along. 

WDYT team? @srindom @zakariaelas @pKorsholm @kasperkristensen 

Also, Kasper, have you noted down the things that are missing in the PR? Cause then we could create tickets for each of those.

kasperkristensen:
@olivermrbl Sounds good, almost all actions (both on actionables and buttons) are missing as they should open some RMA modal. But other than that I think I got everything in there 😄

pKorsholm:
@olivermrbl sounds like a sound plan 😄

PR #278, opened 2022-01-31
Feat: Revamp RMA register swap

pKorsholm:
**What**
- update rma modal for swaps
- sub modal for product selection
- table for return products

image

PR #273, opened 2022-01-27
Feat/revamp rma request return

pKorsholm:
**What** 
- request return revamp

files to review: 
- `src/domain/orders/details/returns/index.js`

*Note: this is dependent on #262*

image

PR #282, opened 2022-02-01
Feat: Revamp rma register claim

pKorsholm:
**What**
- register claim modal
- address submodal

dependent on #273 and #278


srindom:
Claim reasons and return reasons are not the same:
image

Right now we only have these:
```
  const reasonOptions = [
    {
      label: "Missing Item",
      value: "missing_item",
    },
    {
      label: "Wrong Item",
      value: "wrong_item",
    },
    {
      label: "Production Failure",
      value: "production_failure",
    },
  ]
```

srindom:
image

Not sure if we have a module for this but it would be really nice to have this dropdown styled as a single select - the checkbox here makes it seem like you can select multiple reasons + it stays open after you have selected a value which is a bit annoying

srindom (changes requested):
**Country dropdown here**
image

---

**Incorrect formatting here**
image


Last thing and we are ready to go -> please change the country field to a select and make sure that the custom address is displayed correctly. I think the issue is that the country_code is not forwarded between screens correctly.

PR #283, opened 2022-02-01
feat: revamp add collection

kasperkristensen:
**What**

- Adds new create collection modal
- Adds component to add metadata to an entity

Files to review
- `templates/add-collection-modal`
- `organisms/add-metadata`
- `domains/collections/index`

image


src/domain/collections/index.js
```diff
@@ -18,9 +18,51 @@ import {
...
 const CollectionsIndex = () => {
+  const [showNew, setShowNew] = useState(false)
+  const createCollection = useAdminCreateCollection()
+  const toaster = useToaster()
+  const { collections, isLoading, refresh, total_count } = useMedusa(
+    "collections",
+    {
+      search: `?${qs.stringify(filtersOnLoad)}`,
+    }
+  )
```
olivermrbl, line 35:
```suggestion
  const { collections, isLoading, count } = useAdminCollections()
```

src/domain/collections/index.js
```diff
@@ -18,9 +18,51 @@ import {
...
+        }
+      }, {})
+
+      payload.metadata = metadata
+    }
+
+    createCollection.mutate(payload, {
+      onSuccess: () => {
+        toaster("Collection created", "success")
+        refresh()
```
olivermrbl, line 57: Should not be necessary, if we use the hook

olivermrbl (commented):
Great work @kasperkristensen! LGTM, but just had a comment. Let me know 👍

src/domain/collections/index.js
kasperkristensen: Agree but the collection view is being moved to the product overview in another PR and this domain will be deleted so this is just temporary 😄

PR #284, opened 2022-02-01
feat: Login page

olivermrbl:
Adds new login page as well as a new input molecule.
Screenshot 2022-01-31 at 20 28 03
Screenshot 2022-01-31 at 20 27 59


src/components/fundamentals/icons/medusa-icon/index.tsx
```diff
@@ -0,0 +1,22 @@
+import React from "react"
+import IconProps from "../types/icon-type"
+
+const MedusaIcon: React.FC<IconProps> = ({ size = "16", ...attributes }) => {
```
kasperkristensen, line 4: very nit:

the size prop is not used for anything here. This is properly fine since we aren't using it anywhere else atm. But we could do something like this so we can use it in different size at a later time:

```suggestion
const MedusaIcon: React.FC<IconProps> = ({ size = "48", ...attributes }) => {
  const width = +size * 0.9375 // width relative to height (from size prop)
  return (
    <svg
      width={width}
      height={size}
      viewBox="0 0 45 48"
      fill="none"
      xmlns="http://www.w3.org/2000/svg"
      {...attributes}
    >
```

src/components/molecules/input-signin/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+    }: InputProps,
+    ref
+  ) => {
+    const inputRef = useRef(null)
+
+    useImperativeHandle(ref, () => inputRef.current)
+
+    return (
+      <input
+        className={
```
kasperkristensen, line 36:
```suggestion
        className={clsx(
            "outline-none outline-0 remove-number-spinner leading-base",
            "w-[320px] h-[48px] py-3 px-4 mt-4 border rounded-rounded",
            "bg-grey-5 inter-base-regular placeholder:text-grey-40",
            "focus-within:shadow-input focus-within:border-violet-60",
            className
          )}
```

kasperkristensen (approved):
Looks great! 🚀  One small change to the signin input as the style is not applied currently due to clsx being used wrong. But otherwise LGTM

src/components/molecules/input-signin/index.tsx
olivermrbl: Well spotted - thanks! 👍

PR #285, opened 2022-02-01
feat: revamp product overview

zakariaelas:
### What 

- Implements the tile view for products

### How

- Support empty titles for `StatusIndicator`
- Adds an optional `triggerClass` props for `Actionables` to support adding custom classes for the trigger
- Adds an optional `open` prop for `Modal` (useful for controlled modals)
- Extracts the `handleCopyProduct` functionality into a re-usable `useCopyProduct` hook which is used in both the list and tile views
- Extracts the `getProductStatusVariant` method into `src/utils/`
- Adds a `useProductActions` hook which is used to return the actions for each product. This hook can be useful to the list view once we have a `ProductRow` component @pKorsholm 
- Adds a `useImperativeDialog` modal which is used to imperatively invoke a "danger" prompt (will add a few stories later)

### Note

- After a brief discussion with @olivermrbl , we've decided not to include the column headers in the tile view for now until we have support for server-side sorting. 

### Depends on
- #268


src/components/templates/product-table/index.tsx
```diff
@@ -520,6 +450,7 @@ const ProductTable: React.FC<ProductTableProps> = () => {
       },
     ]
   }
+  console.log({ products })
```
olivermrbl, line 453: rm

src/components/templates/product-table/use-copy-product.ts
```diff
@@ -0,0 +1,73 @@
...
+      copy.tags = product.tags.map(({ id, value }) => ({ id, value }))
+    }
+
+    if (product.thumbnail) {
+      copy.thumbnail = product.thumbnail
+    }
+
+    await createProduct.mutateAsync(copy, {
+      onSuccess: ({ product }) => {
+        navigate(`/a/products/${product.id}`)
```
olivermrbl, line 61: It seems to not be navigating to the copied product on success and nor does it display the success message. Are you experiencing the same @zakariaelas?

olivermrbl: But it does create the product. Is it perhaps because we are using `async / await` ?

olivermrbl (commented):
Super nice work @zakariaelas! Really like the approach of using hooks 👍 

Added a couple of comments.

src/components/templates/product-table/overview.tsx
```diff
@@ -0,0 +1,81 @@
...
+          className={clsx("rounded-base inline-block absolute top-2 right-2")}
+        >
+          <Actionables
+            actions={getActions(product)}
+            triggerClass="hidden-actions group-hover:opacity-100 focus-within:opacity-100 opacity-0 bg-grey-0"
+          />
+        </div>
+        <Link to={`${product.id}`}>
+          <img
+            className="min-h-[230px] block object-fill rounded-rounded"
```
pKorsholm, line 58: should we here use `object-cover` instead of `object-fill` to avoid thumbnails getting squashed out of proportion: 
image

src/components/templates/product-table/index.tsx
```diff
@@ -548,41 +479,52 @@ const ProductTable: React.FC<ProductTableProps> = () => {
...
-                  </Table.Row>
-                )
-              })}
-            </Table.Body>
+                </Table.Head>
+                <Table.Body {...getTableBodyProps()}>
+                  {rows.map((row) => {
+                    prepareRow(row)
+                    return (
+                      <Table.Row
```
pKorsholm, line 502: Can we extract this to use the same hooks as the product tile in a component as below?

```javascript 
const ProductRow = ({ row }) => {
  const product = row.original
  const { actions } = useProductActions(product)

  return (
    <Table.Row
      color={"inherit"}
      linkTo={`/a/products/${product.id}`}
      actions={actions}
      {...row.getRowProps()}
    >
      {row.cells.map((cell, index) => {
        return (
          <Table.Cell {...cell.getCellProps()}>
            {cell.render("Cell", { index })}
          </Table.Cell>
        )
      })}
    </Table.Row>
  )
}
```

pKorsholm (commented):
I love it, it looks so nice and the hooks are very elegant

one thing that would be very cool would be to save the state of the product table so when we go back to the product table it doesn't necessarily show the list as default but instead saves the state, I dont know if we should make a provider for it or save it in some other way, but it would be nice to have 😄

src/components/templates/product-table/overview.tsx
zakariaelas: it's just that some pictures will appear as cropped, but I think `object-cover` looks better than `object-fill`

src/components/templates/product-table/use-copy-product.ts
zakariaelas: sorry I missed that :sweat_smile: . Should be fixed now :+1:

olivermrbl (approved):
LGTM! Great work @zakariaelas 🚀

srindom:
Still some shakiness around changing between views and setting limits; but maybe we should just merge for now and fix later?

PR #288, opened 2022-02-01
fix: strip denominations from indexId

zakariaelas:
### What
- strip denominations from `indexId` when submitting 

@olivermrbl dammit I missed this bad boi :disappointed:


olivermrbl (approved):
Ah great, missed that too 😅 Feel free to merge @zakariaelas

PR #290, opened 2022-02-02
Feat: Revamp draggable row

pKorsholm:
Base implementation of a draggable table, without dropping working due to how rapidly it updates state

also updating camelcase on some svgs

files to review: 
- `src/components/templates/draggable-table.tsx`


src/components/templates/draggable-table.tsx
```diff
@@ -0,0 +1,208 @@
...
+      update(records, {
+        $splice: [
+          [dragIndex, 1],
+          [hoverIndex, 0, dragRecord],
+        ],
+      })
+    )
+  }
+
+  const debouncedMoveRow = useMemo(() => _.debounce(moveRow, 100), [])
```
srindom, line 66:
```suggestion
  const debouncedMoveRow = useMemo(() => debounce(moveRow, 100), [])
```

srindom (approved):
Lodash is notoriously heavy so better to only import individual functions to allow proper tree-shaking.

Code looks good otherwise; I assume that it can be seen in action somewhere in a later PR?

zakariaelas (approved):
Nice work Philip! Looking forward to seeing this in action :star_struck:

pKorsholm:
@zakariaelas @srindom I just pushed a storybook story with the draggable table 😄  feel free to give it a shot

PR #291, opened 2022-02-02
feat: Draft Order list overview

olivermrbl:
Adds draft order list over

Includes:
- Table
- Hook for columns
- Hook for filters
- Added "tabs" to switch between orders and draft orders


src/components/templates/draft-order-table/index.tsx
```diff
@@ -0,0 +1,164 @@
...
+  const [numPages, setNumPages] = useState(0)
+
+  const { draft_orders, isLoading, isRefetching, count } = useAdminDraftOrders(
+    queryObject
+  )
+
+  useEffect(() => {
+    const controlledPageCount = Math.ceil(count! / queryObject.limit)
+    setNumPages(controlledPageCount)
+  }, [draft_orders])
```
srindom, line 37:
```suggestion
  }, [count, queryObject])
```

src/components/templates/draft-order-table/use-draft-order-column.tsx
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  const columns = useMemo(
+    () => [
+      {
+        Header: "Draft",
+        accessor: "display_id",
+        Cell: ({ cell: { value }, index }) => (
+          <Table.Cell key={index}>{`#${value}`}</Table.Cell>
+        ),
```
srindom, line 25: Add `pl-2` to this

src/components/templates/draft-order-table/use-draft-order-column.tsx
```diff
@@ -0,0 +1,74 @@
...
+          </Table.Cell>
+        ),
+      },
+      {
+        Header: "Status",
+        accessor: "status",
+        Cell: ({ cell: { value }, index }) => (
+          <Table.Cell key={index}>{decideStatus(value)}</Table.Cell>
+        ),
+      },
```
srindom, line 66: Add `pr-2` to this

can we do this at the table level instead maybe?

srindom (approved):
LGTM - a few minor tweaks that would be nice to include though

src/components/templates/draft-order-table/use-draft-order-column.tsx
olivermrbl: I only think the padding is used on the orders and draft orders tables right now, so will add them to those for now :)

PR #323, opened 2022-02-06
feat: Add images section to Gift Card overview

olivermrbl:


srindom:
I have a Gift Card with this structure:
```
{
  images: [],
  thumbnail: "placeholder.co",
  ...
}
```

The thumbnail is not displayed in this case - the reason is probably that when the gift card was created the thumbnail was set directly instead of first being uploaded as an image and then being selected as the thumbnail. Should we maybe do a "union" thing so it shows unique images across images and thumbnail?

srindom (approved):

PR #328, opened 2022-02-07
fix: polish discounts

zakariaelas:
### What 

- Prevent horizontal overflow when all regions are selected
    - **How**: use css grid
- Enforce minimum value for various number inputs
   - **How**: use `min={0}` or `min={1}`
- Prevent accidental form submission when clicking tooltips or pressing Enter
   - **How**: add `onSubmit={(e) => e.preventDefault()}` to `form` 
- When search results are empty in the discounts table, table pagination should show "0-0 of 0 Discounts" instead of "1-0 of 0 Discounts"
   - **How**: add conditional check when count === 0
- Add support for selecting "00:00" as valid time
   - **How**: fix wrong conditional in `useEffect`
- Add `getDateClassname` method
   - **How**: add missing implementation
- Duplicate typo
   - **How**: "duplicate" instead of "dubplicate"
- Fix number of rows for description textarea
   - **How**: add `rows={8}` prop to `<Textarea />`


olivermrbl (approved):
Great @zakariaelas!

PR #329, opened 2022-02-07
feat: update collection products

kasperkristensen:
**What**
- Adds Medusa request to bulk add/remove products from collections
- Updates the "+ Add Product" button on collection details page to "Edit Products"

**Note**
- Depends on [#1032](https://github.com/medusajs/medusa/pull/1032) in core


src/components/organisms/add-product-modal/index.tsx
```diff
@@ -16,9 +16,14 @@ const AddProductModal: React.FC<AddProductModalProps> = ({
...
-  const handleSelectProduct = () => {
+  const handleSelectProduct = async () => {
     if (selectedProducts.length > 0) {
-      onSubmit(selectedProducts.map((p) => p.id))
+      const allIds = collectionProducts.map((product) => product.id)
+
+      const selectedIds = selectedProducts.map((product) => product.id)
+      const removedIds = allIds.filter((id) => !selectedIds.includes(id))
+
+      await onSubmit(selectedIds, removedIds)
```
srindom, line 26: This is not exactly what we want to do - consider this:
- you have 5 pages worth of products
- you want to add 2 products to the collection; the two products are on the first page
- all the existing products in the collection are on page 4
- you check the two products on page one and click save
- you now only have the 2 products on page 1 in your collection

What we need to do is to keep track of both which products we have checked but also the products which you have unchecked so that we don't do a `filter` when finding the `removedIds` but rather that we have these available in a state variable similar to how `selectedProducts` are controlled

src/services/api.js
```diff
@@ -337,6 +335,16 @@ export default {
...
+
+    addProducts(id, payload) {
+      const path = `/admin/collections/${id}/products/batch`
+      return medusaRequest("POST", path, payload)
+    },
+
+    removeProducts(id, payload) {
+      const path = `/admin/collections/${id}/products/batch`
+      return medusaRequest("DELETE", path, payload)
+    },
```
srindom, line 347: Should we maybe move this to `@medusajs/medusa-js + medusa-react`?

srindom (changes requested):
Small changes - I think we should consider getting the endpoints into `@medusajs/medusa-js + medusa-react` but am happy to approve as is, but let's then open an issue on getting that done later :)

src/services/api.js
srindom: Have created an issue on this: https://github.com/medusajs/medusa/issues/1089

srindom (approved):
LGTM!

PR #335, opened 2022-02-08
Fix: dropdown overflow

pKorsholm:
**What** 
- fix overflow issue with dropdown

**How** 
- switch dropdown component from `react-multi-select-component` to `react-select` (which has more users, stars and downloads in general)
- `react-select` supports native portalling of the dropdown which enables us to show it as an overlay instead, which `react-multi-select-component` doesn't have
- add `ModalContext` which exposes a ref to the modal portal

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "react-json-view": "^1.19.1",
     "react-multi-select-component": "^4.1.15",
     "react-popper": "^2.2.5",
-    "react-select": "^4.1.0",
+    "react-select": "^5.2.2",
     "react-table": "^7.7.0",
     "react-toast-notifications": "^2.4.0",
     "react-tooltip": "^4.2.10",
--- a/src/assets/styles/global.css
+++ b/src/assets/styles/global.css
@@ -163,54 +163,57 @@
   .radio-outer-ring > span[data-state="checked"] {
     @apply rounded-circle shadow-[#7C3AED] shadow-[0_0_0_2px];
   }
+}
 
-  .multiselect-styling {
-    @apply p-0 -mx-3 mt-1 outline-0 cursor-text;
+@layer components {
+  .react-select-container {
+    @apply p-0 -mx-3 border-0 mb-1 cursor-text h-6;
 
-    .dropdown-container {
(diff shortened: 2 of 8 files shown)


kasperkristensen (commented):
Very cool! Howver, I noticed two issues. 

You are able to take the dropdown for a walk when scrolling 😅 

https://user-images.githubusercontent.com/45367945/154028598-e48e095b-f1b7-416b-821f-804d4d621c64.mov

And if you open the dropdown using the chevron then search field is not focused and the dropdown will have a different position then when it the search fields is focussed:

https://user-images.githubusercontent.com/45367945/154030290-35dfb1f0-f970-4a45-8e32-c4ea8bbea851.mov

kasperkristensen:
Ahh cool, this brings it very close to being perfect. Closing it on scroll does result in some weird behaviour if the select is very close to the bottom of the viewable part of a scrollable div. This scrolls the window a bit and then closes the Select immediately 

https://user-images.githubusercontent.com/45367945/154089654-f49a177d-27c7-47c1-9637-7f2635a64502.mov

olivermrbl:
Also getting strange behaviour when trying to open other dropdowns, when one is already open. It seems, that in order for us to open a different dropdown and close the currently open one, we have to either scroll to close it, or close it by clicking on it.

See attached video for reference.

https://user-images.githubusercontent.com/59018053/154245489-870fc2a4-56be-4b13-ad59-cd0dd97cee7e.mov

pKorsholm:
I think we are closing in 😄  changed up the open/close monitoring in favor of the built in controls and instead now monitor for focus and blur. 

I fixed the issue with 'jump scrolling using a debounce such that scrolling in the first 50 ms after opening the dropdown will not close it. If you have a better suggestion please let me know, issue is that scroll events when opening the dropdown and regular scrolling are indistinguishable. 


https://user-images.githubusercontent.com/88927411/154293633-c9201b32-1ea5-42f5-8ef6-c5b198b5e0a6.mov

olivermrbl:
To somewhat re-use @kasperkristensen's epic phrase from above, I _can't_ take the dropdown for a walk, if I resize the window 😅 

image

PR #338, opened 2022-02-09
fix: various bugs on Gift Card pages

kasperkristensen:
**What**

- [x] Deleting a gift card displays two toasters
- [x] Error message in "Update balance modal" when updated balance in higher than value (currently we just disable button)
- [x] Subtitle to Gift Cards page should be "Manage the Gift Cards of your Medusa store" 
- [x] Edit -> Cancel -> Displays toaster
- [x] Email validation on input field
- [ ] Fix can't update a custom gift card to a balance of 0
- [x] Adds a " - " in the GiftCardTable Order column when no order is present
- [x] Add status update to dropdown menu as well
- [x] Fix so tooltips click no longer triggers re-renders
- [x] Make Gift Card name required and trim
- [x] Updates loading state so title "Undefined" is no longer displayed while a gift card is loading

**Note**
- Updating a balance to 0, will be fixed in a PR to the core, as it is currently not supported.


olivermrbl:
As per request from design team, I've updated gift card to Gift Card

olivermrbl (approved):
LGTM, great work!

PR #341, opened 2022-02-09
feat: revamp toaster

kasperkristensen:
**What**

- Revamps toaster design

ezgif-2-a62ea3e6c6

**Note**

- Since the new design has both a message and a title, I have updated every instance where toaster is used to account for this change. This means that a lot of files have been changed.
  - To review this PR you should just focus on the files `atoms/toaster`, `hooks/use-toaster`, `gatsby-browser`.
  - For now I have just entered `Success` or `Error` for the titles. We should definitely change this, but I think we need to have a discussion of how to format our success/error/warning and info messages.
  - Also the animation is more smooth then what appears on the gif 😅


src/components/atoms/toaster/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+      >
+        <div>{getIcon(appearance)}</div>
+        <div className="flex flex-col ml-small mr-base gap-y-2xsmall flex-grow">
+          {children}
+        </div>
+        <div>
+          {/* @ts-ignore:next-line */}
+          <div role="button" onClick={onDismiss} className="cursor-pointer">
+            <CrossIcon size={ICON_SIZE} className="text-grey-40" />
+          </div>
```
zakariaelas, line 75: just curious to know why a `div` instead of `button`? Tried using a button and couldn't notice any differences (visually), but might have missed it :smile:

src/components/atoms/toaster/toaster.stories.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+export default {
+  title: "Atoms/Toaster",
+  component: Toaster.Content,
+} as ComponentMeta<typeof Toaster.Content>
+
+const Template: ComponentStory<typeof Toaster.Content> = (args) => (
+  <div className="flex items-start bg-grey-0 p-base border border-grey-20 rounded-rounded shadow-toaster w-[380px]">
+    <Toaster.Content {...args} />
+  </div>
```
zakariaelas, line 13: nit: not sure if this is intended, but this is how the stories look like:
image

zakariaelas (approved):
awesome work @kasperkristensen ! It looks reaaaaally good and I loooove the animations :heart_eyes:. Left a tiny comment on the stories

olivermrbl (approved):
LGTM, great work @kasperkristensen !

PR #345, opened 2022-02-10
fix: Mark swap + claim fulfillments as shipped

olivermrbl:
Additionally, refactors cancel fulfillment into fulfillment component and adds the use of imperative dialogs instead of separate modals for each delete / remove operation.


pKorsholm (approved):
lgtm 😄

PR #346, opened 2022-02-10
fix: scrollbar styling

zakariaelas:
### What

- adds a new scrollbar design
- changes occurences of `overflow-y-scroll` to overflow-y-auto` to only bring up the scrollbar when needed


olivermrbl (approved):
LGTM, very clean!

PR #347, opened 2022-02-10
feat: add global search

zakariaelas:
### What

- add global search
- implements keyboard navigation through the `useKeyboardNavigationList` hook

**Known issues**

- when providing a query to the `/products` endpoint, the limit is ignored, so you might get a lot of results based on your query, but it should be fine once we fix the issue in the core
- when using keyboard navigation to reach results at the bottom of the list, the modal does not scroll down (will fix later)


src/components/molecules/search-bar.tsx
```diff
@@ -1,76 +1,50 @@
...
-            onKeyDown={onKeyDown}
-            onChange={handleChange}
-            ref={searchRef}
-            placeholder="Search anything..."
-          />
-        </>
-      ) : null}
-    </div>
+    <>
+      <button
```
pKorsholm, line 34: Should this be wider you think? 

It feels a bit off when you can only click the small text and symbol in the upper left, we could definitely also leave it as is without any issues.

pKorsholm (commented):
this is beautiful! words cant even describe how cool this is, amazing 😍 

One thing came to mind though, should removing all text reset the search? as it is right now it leaves the last search on screen.

srindom:
Also Ludvig added designs for an order section :p

src/components/molecules/search-bar.tsx
zakariaelas: totally! good catch :+1:

zakariaelas:
@srindom @pKorsholm thanks for the comments! Made some adjustments for you to check out :smile:. 

One thing to note here: the clear search tooltip is a bit different from Ludvig's design (see [here](https://www.figma.com/file/dlraOnQvQnrHQL0A7Ysbrs/Web-App-2.0?node-id=2789%3A11520)). The difference is that I have omitted the "ESC" key from the tooltip because pressing ESC when the tooltip is showing will close down the entire modal and not just the tooltip. I'm not entirely sure if that's intended behavior, but that's how it plays out with Radix's components right now. Thought it made more sense to not give this one too much thought and just omit it :sweat_smile:

olivermrbl:
Loving it, great work @zakariaelas! To the point about the tooltip, I think the current behavior is fine, and it seems pretty intuitive to me that `esc` closes the modal. 

If @srindom and @pKorsholm have no further comments to their previous requested changes, we should be ready to merge 👍

src/components/templates/search-modal/index.tsx
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  const { orders, isFetching: isFetchingOrders } = useAdminOrders(
+    {
+      display_id: query,
+      limit: 5,
+      offset: 0,
+    },
+    { enabled: !!query, keepPreviousData: true }
+  )
```
srindom, line 41: Don't know if I have confused things a bit but let me try to explain two stories that I think we should solve for when querying for orders:

## Story 1
- I am a customer service user and I get an email from a customer who sends me her order id (i.e. `order.display_id`) - I copy the order id 
- If I do `cmd + k` then `cmd + v` that should take me straight to the order - I don't necessarily want to wait for all the other queries to complete.

## Story 2
- I am receiving a call from a customer and they don't have their order number but they give me their email
- I want to do `cmd + k` and start typing their email; I want to see their most recent order or alternatively just their customer profile

Right now the order query here doesn't solve for Story 2; my thoughts on how we could accomplish both stories would be this:

Get orders by the free text search i.e.:
```suggestion
  const { orders, isFetching: isFetchingOrders } = useAdminOrders(
    {
      q: query,
      limit: 5,
      offset: 0,
    },
    { enabled: !!query, keepPreviousData: true }
  )
```

Additionally add another query like this one:
```js
  const { orders: directHitOrders, isFetching: isFetchingOrders } = useAdminOrders(
    {
      display_id: query,
      limit: 5,
      offset: 0,
    },
    { enabled: !!query, keepPreviousData: true }
  )
```

If we have a hit on i.e. `directHitOrders.length === 1 && directHitOrders[0].display_id === query` then take us straight to the order else ignore `directHitOrders`

--- 

The stuff above was what my initial thought was what I am realizing now, however, is that if we just go with the suggestion above my "cursor" will already be on the order I want to go to so all I would have to do would be to hit `enter` and I would get there - I think that is a fair flow to stick with so no need to add the additional query. 

**So in conclusion** let's allow users to also search orders by email i.e. change `display_id` to `q`. I realize this is a lot of talk for a simple fix but just wanted to open-source my thinking :D

src/components/molecules/search-bar.tsx
```diff
@@ -1,76 +1,50 @@
...
-      setQuery(e.target.value)
-    }
+  const closeModal = () => {
+    setShowSearchModal(false)
   }
 
-  useHotkeys("cmd+k", hotKeyFocus, {}, [searchRef])
-  useHotkeys("ctrl+k", hotKeyFocus, {}, [searchRef])
+  useHotkeys("cmd+k", toggleSearch, {}, [])
+  useHotkeys("ctrl+k", toggleSearch, {}, [])
```
srindom, line 22: #vimlyfe

```suggestion
  useHotkeys("cmd+k", toggleSearch, {}, [])
  useHotkeys("ctrl+k", toggleSearch, {}, [])
  useHotkeys("/", toggleSearch, {}, [])
```

srindom: Also have you checked if `useHotkeys` works with `Meta+k`? I think that might be a way to capture both `ctrl` and `cmd`

srindom (changes requested):
Amazing work love this feature! Have some very small tweaks :))

src/components/molecules/search-bar.tsx
zakariaelas: I don't think it does for now. Btw, according to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/metaKey) it appears to me that Ctrl !== Meta so not entirely sure if it's a common thing libraries implement :sweat_smile:

src/components/templates/search-modal/index.tsx
zakariaelas: thanks for taking the time to explain @srindom! I also think the flow we have rn should be good enough, although what you mentioned in Story 1 could be super useful!

Although one edge case I think we should be a tiny bit careful about is one where the order number conflicts with something like a discount code. For example, user types in 10 initially to look for a `DISCOUNT10`, if it turns out that there is an order with display_id 10, then we'd always be prioritizing the order, even though the user is looking for a discount. This is probably a bit of a stretch because the probability might be very small, but could still happen I guess :sweat_smile:

srindom (approved):
LGTM!

PR #359, opened 2022-02-14
Adding Metadata Fields to the Variant

jaruesink:
image


src/domain/products/details/index.js
```diff
@@ -50,9 +50,8 @@ const ProductDetail = ({ id }) => {
     })
   }
 
-  if (didFail) {
-    return <NotFound />
-  }
+  if (didFail) return <NotFound />
+  if (!product) return <Spinner />
```
srindom, line 54: We use `eslint curly` so please 

```suggestion
  if (didFail) {
    return <NotFound />
  }
  if (!product) {
    return <Spinner />
  }
```

Also it seems that `Spinner` is not imported

src/components/organisms/metadata/index.tsx
```diff
@@ -61,7 +61,7 @@ const Metadata: React.FC<AddMetadataProps> = ({
       <div className="flex flex-col mt-base gap-y-base">
         {localData.map((field, index) => {
           return (
-            <DeletableElement key={index} onDelete={deleteKeyPair(index)}>
+            <DeletableElement key={field.key} onDelete={deleteKeyPair(index)}>
```
srindom, line 64: Not an expert on this, but I think we need to use a constant value for the key prop. The `field.key` value might change when editing the metadata so maybe this is not that appropriate?

src/domain/products/details/variants/variant-editor.tsx
```diff
@@ -23,6 +24,12 @@ const VariantEditor = ({ variant, onSubmit, onCancel }) => {
   const { store, isLoading } = useAdminStore()
   const [currencyOptions, setCurrencyOptions] = useState([])
   const [prices, setPrices] = useState(variant.prices)
+  const [metadata, setMetadata] = useState(
+    Object.keys(variant.metadata).map((key) => ({
+      key,
+      value: variant.metadata[key],
+    }))
+  )
```
srindom, line 32: Why not just:
```suggestion
  const [metadata, setMetadata] = useState(variant.metadata)
```

srindom (changes requested):
This is great!

package.json
```diff
@@ -24,7 +24,6 @@
     "@radix-ui/react-tooltip": "^0.1.6",
```
srindom, line 24: any reason why you have made these updates? - had issues when starting the dev server as react was missing.

Don't think we need to include this file in the PR :)

src/domain/products/details/index.js
jaruesink: oh I wonder why vscode didn't warn me about that, I'll update it!

src/components/organisms/metadata/index.tsx
jaruesink: @srindom using the index here had a bug where if you deleted an item, then it would show the wrong values. This seemed to fix that issue.

src/domain/products/details/variants/variant-editor.tsx
jaruesink: @srindom the metadata form wants a shape of `{key: string, value: string}[]`, but the metadata is just type `{[x:string]: string}`

jaruesink: @srindom the metadata form wants a shape of `{key: string, value: string}[]`, but the metadata is just type `{[x:string]: string}`

package.json
jaruesink: @srindom oh shoot, I was trying to clean things up late at night. let me fix this.

package.json
```diff
@@ -24,7 +24,6 @@
...
     "@rebass/forms": "^4.0.6",
     "@svgr/webpack": "^5.4.0",
-    "@tailwindcss/line-clamp": "^0.3.1",
```
jaruesink, line 27: this was included twice, so I left the one in the devDependencies

jaruesink:
@srindom I fixed the merge conflicts to keep this up to date. Still working well for us.

PR #360, opened 2022-02-14
fix: missing product details

zakariaelas:
### What

- support publishing/unpublishing products through the edit products page
- support setting whether a product should be discountable or non-discountable


srindom (approved):
LGTM!

PR #362, opened 2022-02-14
feat: Detect changes on product new and edit

kasperkristensen:
**What**
- Adds `SaveNotification`'s too `edit`/`new` product routes.
- Show `Save/Discard` notification when dirty state is detected in form

**How**
- Uses new `NotificationProvider` that should be used as a child of a form provider.
- Takes a `onReset` function to reset the form on discard changes, a `onSave` that can either be a single function, or multiple function (as an example on `new` page there is a `Save as draft` and `Save and publish`), and `additionalDirtyStates` as param. The `additionalDirtyStates` should contain other states that show that the form is dirty that useForm will not pick up on. An example of this is if you change the order of images of products.

[See video](https://user-images.githubusercontent.com/45367945/155723720-a015d790-f4cc-4a24-8479-29637b52ebb1.mov)

**Note**
- Design (especially color scheme) might change, as well as a design for how to handle the notification overlapping input fields.


pKorsholm (commented):
LGTM

srindom:
@kasperkristensen please resolve conflicts so we can merge :)

PR #363, opened 2022-02-14
Fix: file upload error with no files uploaded

### LINKED ISSUE #350: Cannot Add Product |  Could not resolve 'fileService'.
Not sure if this belongs in the admin-panel or in the core medusa repo, but I'm having a problem where I basically can't add new products to the admin UI as every time I add one I get the following error:
```
Resolution path: fileService
AwilixResolutionError: Could not resolve 'fileService'.
```

new-product-fail

According to[ this core issue ](https://github.com/medusajs/medusa/issues/562) the problem is that the admin site is attempting to upload the images to a fileserver. This makes sense but:
1. There are no images being uploaded, so it shouldn't need to use the File Service
2. The error message should be more friendly. It should say something like "Cannot upload image without File Service Plugin".
3. If you do need to set up the File Service before using the admin panel, you should really add more detailed documentation on that, at least a link at the bottom of the quickstart.

Steps to reproduce:
1. Follow the quickstart guide in the main `readme.md`
2. Try to add a new product, only filling out the name and not uploading any files.
3. Attempt to add the new product

pKorsholm:
**What** 
- only upload images if any images are present, otherwise dont send the request to the server 

fixes #350

### DIFF
--- a/src/domain/products/new.tsx
+++ b/src/domain/products/new.tsx
@@ -21,18 +21,19 @@ const NewProductPage = () => {
     const images = data.images
       .filter((img) => img.url.startsWith("blob"))
       .map((img) => img.nativeFile)
-    const uploadedImgs = await Medusa.uploads
-      .create(images)
-      .then(({ data }) => {
-        const uploaded = data.uploads.map(({ url }) => url)
-        return uploaded
-      })
-    const newData = {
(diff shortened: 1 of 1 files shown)


olivermrbl (approved):
LGTM!

PR #365, opened 2022-02-15
Fix: error on invite page

pKorsholm:
**What**
- move the medusa logo into the invite box same way as on login
- Fix the following error when going to `/invite` with or without token using [this approach](https://www.gatsbyjs.com/docs/reference/release-notes/migrating-from-v2-to-v3/#process-is-not-defined) outlined by gatsby

image


kasperkristensen:
Think we can close this as it is fixed in #376 ✌🏻

olivermrbl:
Part of #376

PR #366, opened 2022-02-15
feat: variant grid on product details and creation

zakariaelas:
### What

- adds variant grid to product details and creation


olivermrbl (commented):
Would it make sense to remove the Pricing card on simple products and just force the user to use the edit variant modal? @srindom @zakariaelas  

image

srindom:
> Would it make sense to remove the Pricing card on simple products and just force the user to use the edit variant modal? @srindom @zakariaelas
> 
> image

Yes! I think that is a great idea - in the future we could introduce some sort of identifier that would make it more verbose that a product is in fact a simple product. I think the guessing game of saying that if there is only one variant it is a simple product might not play out too well.

srindom (changes requested):
Let's kill the price section and then we are good to go! :)

zakariaelas:
sounds good guys! To re-iterate, we're getting rid of the "simple product" vs "product with variants" views, and we're just going with a more generic view which will exclude the pricing card and include the variants card (+ eveything else), is that correct? @olivermrbl @srindom

srindom:
@zakariaelas yes that is correct - the exception, however, is when creating a new product; I think that it is pretty nice that you are not forced to come up with an option and option value when you don't need it.

Also I just spotted that when creating a product with variants there is an issue with parsing the form data:
image

zakariaelas:
@srindom couldn't get to reproduce the error, but after making some changes, tried with similar inputs and worked fine (un/fortunately). Let me know if you can trigger the error again :sweat_smile:

srindom (approved):

PR #368, opened 2022-02-15
Fix: Tag input blur case

fPolic:
**What**

Fix of the tag input `onBlur` behavior.

**Why**

If the user hasn't added a comma as the last input character the last tag would not register.

**How**

Updated the `handleBlur` method to also call the `handleAddValue`.

### DIFF
--- a/src/components/molecules/tag-input/index.tsx
+++ b/src/components/molecules/tag-input/index.tsx
@@ -132,7 +132,12 @@ const TagInput: React.FC<TagInputProps> = ({
   }
 
   const handleBlur = (e) => {
+    const value = inputRef?.current?.value
     setHighlighted(-1)
+
+    if (value) {
+      handleAddValue(value)
+    }
   }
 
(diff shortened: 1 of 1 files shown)


pKorsholm (approved):
LGTM 😄 

one thing we probably should create an issue on is that a string of only spaces can also be created as a tag, not just here but generally and we should handle that in `handleAddValue` if that is not the intended functionality. Also we could consider trimming the strings to ensure that trailing or leading spaces are removed. Thoughts @olivermrbl and @srindom?

srindom:
> LGTM 😄
> 
> one thing we probably should create an issue on is that a string of only spaces can also be created as a tag, not just here but generally and we should handle that in `handleAddValue` if that is not the intended functionality. Also we could consider trimming the strings to ensure that trailing or leading spaces are removed. Thoughts @olivermrbl and @srindom?

This is a great point - would probably also be a good idea to trim the strings before creating the tags

srindom (approved):

PR #374, opened 2022-02-15
fix: enable separate email updates for order overview

fPolic:
**What**

 - Added a dropdown option that allows the email edit
 - Added the email modal

**Why**

 - Changes enable a merchant to edit the email on an order separately from the "Edit shipping/billing address" modal


kasperkristensen (approved):
Great job Frane! LGTM 😄

PR #379, opened 2022-02-16
fix: : swap receive and payment

kasperkristensen:
**What**

- Fixes receive return on swaps
- Adds actionable to capture payments on swaps


olivermrbl (approved):
LGTM!

PR #380, opened 2022-02-16
Fix: dropdown overflow

pKorsholm:
**What**
- fix overflow issue with dropdown

**How**

- switch `dropdown` component from `react-multi-select-component` to `react-select` (which has more users, stars and downloads in general)
- `react-select` supports native portalling of the dropdown which enables us to show it as an overlay instead, which `react-multi-select-component` doesn't have
- add `ModalContext` which exposes a ref to the modal portal

this pr replaces #335 because of the pr base and excessive merge conflicts

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
     "react-json-view": "^1.19.1",
     "react-multi-select-component": "^4.1.15",
     "react-popper": "^2.2.5",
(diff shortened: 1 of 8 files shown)


olivermrbl (approved):
LGTM! Feel free to merge 👍

PR #385, opened 2022-02-21
fix: broken gradient on safari

### LINKED ISSUE #352: BodyCard gradient on Safari is off when scrolling
image

srindom:
Closes #352.

Fixes a bug that renders the "fading" gradient effect in the body card in a weird way on Safari.

Apparently, [Safari treats `transparent` differently](https://stackoverflow.com/questions/38391457/linear-gradient-to-transparent-bug-in-latest-safari#:~:text=For%20most%20browsers,along%20the%20gradient.) than other browsers so that gradients that use the `transparent` keyword look weird. 

Also added `rounded-t-rounded` to the overlay so that the corners don't overflow the body card.

### DIFF
--- a/src/components/organisms/body-card.tsx
+++ b/src/components/organisms/body-card.tsx
@@ -42,7 +42,7 @@ const BodyCard: React.FC<BodyCardProps> = ({
       {...rest}
     >
       {isScrolled && (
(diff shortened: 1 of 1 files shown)


olivermrbl (approved):
LGTM!

PR #391, opened 2022-02-23
fix: toaster breaking layout in prod

kasperkristensen:
**What**
- Moved toaster to `layout` as it does not play nice with the `gatsby-focus-wrapper`
- Also removed the use of toasters on the reset-password page as it was not inline with how we handle errors on other "login" pages
- Also changed `LoginLayout` to `tsx` and moved it to templates

**Testing**
- Tested by building and serving the output


olivermrbl (approved):
LGTM!

PR #396, opened 2022-02-23
feat: detect changes discount form

kasperkristensen:
**What**

- Implements detect changes and show notification pattern in discounts
- revamps domain to use same "form pattern" as used in the product domain
- Adds optional `hideCurrency` prop to `CurrencyInput` so InputCurrency can be used without showing the currency selector.

**Note**
- Dependant on #362


pKorsholm (commented):
LGTM 

I think we should get this in before we start work on promotions because the discount-form will help out tremendously in creating the new promotions page.

srindom:
@kasperkristensen can you please resolve conflicts so we can merge? :)

srindom (approved):

PR #398, opened 2022-02-24
fix: exchange payment link

fPolic:
**What**
 - use `store.swap_link_template` instead of `store.payment_link_template` to generate the payment link used for the exchange item
 
 **Why**
  - the user was unable to copy the payment link and view the chart id


olivermrbl (approved):
Great!

PR #399, opened 2022-02-24
fix: empty and duplicated filters + page count

zakariaelas:
### What

- [x] Trim "Save filter" to avoid blank names
- [x] Prevent duplicate saved filters
- [x] Empty products table shows page "1 of 0"


olivermrbl (approved):
LGTM!

PR #400, opened 2022-02-24
fix: collections breadcrumb would take you to /collections instead of…

kasperkristensen:
**What**
- Fixes breadcrumb on edit collection page as it would take you to `/collections` instead of `/products?view=collections`
- removes `/collections` route
- converts `collections/index` to `ts`


olivermrbl (approved):
LGTM! 👍

PR #401, opened 2022-02-24
chore: Clean up unused components

olivermrbl:
**What**
Remove all unused UI components


kasperkristensen (commented):
LGTM! Just some more files we could clean up 😄 

- We should remove `assets/svg` folder and rename `svg-2.0` to just `svg`. This will require updating the paths in `emoji-picker.css`.
- We should uninstall the following dependencies: 
  - `gatsby-plugin-portal`
  - `prop-types` (will need to remove its usage from `more-horizontal-icon.tsx` and `trash-icon.tsx`)
  - `react-collapse`
  - `react-collapsible`
  - `rebass`
  - `@rebass/forms`
  - I also think we can remove `@popperjs/core`, but I am not sure if it is being used under the hood by our tooltip implementation. But maybe someone else knows.

olivermrbl:
> We should uninstall the following dependencies:

Unused and deprecated dependencies will be removed in #406 👍

adrien2p (approved):
LGTM! All those components have been replaced or they are just not needed anymore?

olivermrbl:
> All those components have been replaced or they are just not needed anymore?

A mix of both 👍 @adrien2p

adrien2p:
Wow it will make the admin much more light 😂

pKorsholm (approved):
This looks great! 

I think we can remove `components/table` it doesn’t seem to be referenced anywhere. We should also be able to remove `components/not-found.js` as it doesn't seem to be referenced either.

Also we have a couple of components which are used but not grouped according to the components structure of the folder, we could move it here or create a new ticket. up to you 😄 
- move `components/declarative-toaster` to atoms 
- move `components/loading-container` to atoms
- move `components/variant-grid`  to templates or molecules

That leaves `seo` and `private-route` which we could throw into one of the folders, however they semantically seem to be different from the rest of the components. 

We could consider moving `private-route` into `a.js` as a functional component that lives there since it has limited usability outside `a.js`. 

I dont have any suggestions for `seo` other than we could move it into atoms or keep it where it is.

olivermrbl:
> Also we have a couple of components which are used but not grouped according to the components structure of the folder, we could move it here or create a new ticket. up to you 😄
> 
> * move `components/declarative-toaster` to atoms
> * move `components/loading-container` to atoms
> * move `components/variant-grid`  to templates or molecules

This is great @pKorsholm - thanks! Have created separate issues for them 👍

PR #404, opened 2022-02-24
fix: product bug hunt fix details

zakariaelas:
### What 
- [x] Trim product name on create (credits to @kasperkristensen)
- [x] Disallow negative values in stock, dimensions, etc.
- [x] Allow strings in Country of Origin field
- [X] Alignment on price rows
- [X] Adding new prices should filter out already used currencies
- [X] Added support for duplicating variants


olivermrbl (commented):
Great work @zakariaelas! 

Not a part of the bug hunt, but spotted a couple of issues with the collections dropdown, that I think we might as well fix now:
1. You can't de-select a collection
2. If you try to de-select a collection and scroll, you are taking the dropdown for a walk

The latter is more related to the dropdown refactor from last sprint, so @pKorsholm might be able to pitch in on that one. And also totally fine with creating a new issue to bring into this sprint, if you find it appropriate 👍

zakariaelas:
@olivermrbl sure thing! Will tackle them right now :smile:

olivermrbl (approved):
LGTM!

PR #420, opened 2022-03-04
fix: currency input width

zakariaelas:
### What

- fixes currency input styling when `size="full"`

### How

- instead of doing `basis-full` we can just use `flex-1` instead. This allows the currency input to take the full width of the container in case the currency input does not have children, otherwise container width will be split equally between both children


olivermrbl (approved):
LGTM!

PR #424, opened 2022-03-10
Fix: Gift card creation

pKorsholm:
**What**
- register names of currency input fields

**Why**
- denominations is not populated in `submit` method causing it to always fail even when denominations are added in the ui

### DIFF
--- a/src/domain/gift-cards/new.tsx
+++ b/src/domain/gift-cards/new.tsx
@@ -46,6 +46,7 @@ const NewGiftCard: React.FC<NewGiftCardProps> = ({ onClose }) => {
 
   const addDenomination = () => {
     const name = `denominations.${denominations.length}`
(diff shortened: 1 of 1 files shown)


bennetfabian:
@pKorsholm What's the reason you closed the PR? Has your approached turned out to not work?

bennetfabian:
Cause at least for me it's still not working. Not even sending out a request to server so it probably is client side.

olivermrbl:
@bennetfabian Branch `develop` should fix your issue. We've cleaned up the gift card domain.

bennetfabian:
> @bennetfabian Branch `develop` should fix your issue. We've cleaned up the gift card domain.

Thanks. Is there a reason why this particular fix hasn't been added to master yet?

PR #428, opened 2022-03-14
fix: free shipping discount

kasperkristensen:
**What**
- Makes it possible to create a `free_shipping` discount.
- Also fixes `location.state` type on `discounts/new`
- also changes "draft" button to "disable"


olivermrbl (approved):
LGTM!

PR #429, opened 2022-03-14
fix: make type createable

kasperkristensen:
**What**

- Makes it possible to create new product types in the "Type" dropdown on products page.
- Always enables search on createable selects

### DIFF
--- a/src/domain/products/product-form/sections/general.tsx
+++ b/src/domain/products/product-form/sections/general.tsx
@@ -1,3 +1,5 @@
+import { useParams } from "@reach/router"
+import { navigate } from "gatsby"
 import {
   useAdminCollections,
   useAdminDeleteProduct,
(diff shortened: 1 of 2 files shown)


src/domain/products/product-form/sections/general.tsx
```diff
@@ -40,6 +46,19 @@ const General = ({ showViewOptions = true, isEdit = false, product }) => {
       value: collection.id,
     })) || []
 
+  const setNewType = (value: string) => {
+    console.log(value)
```
pKorsholm, line 50:
```suggestion

```

pKorsholm (commented):
Looks good! 😄 

One thing, if you create a new type and go to a different product to apply the same type (using menu navigation) the new type isn't shown in the select, however if you refresh it shows up. I think this might be because of the medusa-react cache.

pKorsholm (approved):
LGTM too 😄

PR #431, opened 2022-03-16
feat: customer groups page

fPolic:
**What**
 - new customer groups & customer group details pages
 - `useQueryFilters` hook
 - `useSetSearchParams` hook
 
 **TODO**
  - [x] use https://github.com/medusajs/medusa/pull/1113 for customers list on the details page (new ticket?)
  - [x] "lift" edit logic into a context component
  - [x] fix pagination component breaks if initially `limit == 0` in the URL
  - [x] fix `useSortBy` hook causing infinite renders on query filter change


src/hooks/use-query-filters.ts
```diff
@@ -0,0 +1,217 @@
...
+ */
+function buildQueryObject(
+  state: FilterState,
+  toQuery: Record<string, string | number> = {}
+) {
+  for (const [key, value] of Object.entries(state)) {
+    if (key === "q") {
+      if (typeof value === "string") {
+        if (value) toQuery["q"] = value
+        else delete toQuery["q"]
```
srindom, line 64:
```suggestion
        if (value) {
          toQuery["q"] = value
        } else {
          delete toQuery["q"]
        }
```

src/hooks/use-query-filters.ts
```diff
@@ -0,0 +1,217 @@
...
+    ...ADMIN_DEFAULT_PARAMS,
+    ...defaults,
+  } as FilterState
+
+  if (!queryString) return representation
+
+  const filters = qs.parse(queryString)
+
+  for (const [key, value] of Object.entries(filters)) {
+    if (typeof value !== "string") continue
```
srindom, line 105:
```suggestion
    if (typeof value !== "string") {
      continue
    }
```

src/hooks/use-query-filters.ts
```diff
@@ -0,0 +1,217 @@
...
+    const nextState = { ...state }
+    // TODO: merge and change refs along the `action.path`
+    set(nextState, action.path, action.payload)
+
+    return nextState
+  }
+
+  if (action.type === FilterActionType.SET_QUERY)
+    // if the query term has changed reset offset to 0 also
+    return { ...state, q: action.payload, offset: 0 }
```
srindom, line 141:
```suggestion
  if (action.type === FilterActionType.SET_QUERY) {
    // if the query term has changed reset offset to 0 also
    return { ...state, q: action.payload, offset: 0 }
  }
```

src/hooks/use-query-filters.ts
```diff
@@ -0,0 +1,217 @@
...
+
+  if (action.type === FilterActionType.SET_OFFSET)
+    return { ...state, offset: action.payload }
```
srindom, line 144:
```suggestion
  if (action.type === FilterActionType.SET_OFFSET) {
    return { ...state, offset: action.payload }
  }
```

src/hooks/use-set-search-params.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Effect hook which reflects `queryObject` k/v in the url.
+ */
+function useSetSearchParams(queryObject: Record<string, string | number>) {
+  useEffect(() => {
+    const url = new URL(window.location.href)
+
+    for (let k of url.searchParams.keys()) {
+      if (!(k in queryObject)) url.searchParams.delete(k)
```
srindom, line 11:
```suggestion
      if (!(k in queryObject)) {
        url.searchParams.delete(k)
      }
```

src/components/molecules/customers-groups-summary/index.tsx
```diff
@@ -0,0 +1,31 @@
+import React from "react"
```
srindom, line 1: nit: this is not yet used - is the usage coming in another PR?

src/components/molecules/customers-groups-summary/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+interface P {
+  groups: CustomerGroup[]
+}
+
+/**
+ * Render a summary of groups to which the customer belongs
+ */
+function CustomersGroupsSummary(props: P) {
+  const groups = sortBy(props.groups, "name")
+  if (!groups.length) return null
```
srindom, line 18:
```suggestion
  if (!groups.length) {
    return null
  }
```

src/components/molecules/customers-groups-summary/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+  const left = groups.length - 1
+  const leadName = groups[0].name
```
srindom, line 21: nit: just a suggestion - would it make sense to add a `limit` prop that defines "how many" groups will be displayed in the lead?

And then maybe rename this molecule to something that signifies that this is also usable for other areas not just customer groups. Maybe "LabelSummary", "EllipsisList", "EllipsisLabels" or something like that :D

srindom (changes requested):
Great work mostly just small nits all the functionality is 👌 

We generally don't want ifs without braces so have tried to suggest on all of the occurrences.

I believe we have `curly: all` in our eslint rules so would you mind double-checking that it is configured correctly :)

src/components/molecules/customers-groups-summary/index.tsx
fPolic: It's used in: `src/components/templates/customer-group-table/config.tsx`

srindom: Sorry my bad!

fPolic:
> We generally don't want ifs without braces so have tried to suggest on all of the occurrences.
> 
> I believe we have `curly: all` in our eslint rules so would you mind double-checking that it is configured correctly :)

Switched my editor while working on the feature and eslint wasn't configured properly 😬 . Now should be OK.

src/components/molecules/customers-groups-summary/index.tsx
srindom: Let's leave it as is for now and keep a refactor in mind if we come across the need for a similar component in the future! :)

srindom (approved):
LGTM

PR #452, opened 2022-03-24
feat: `Switch` atom

fPolic:
**What**
- add new `Switch`component using Radix primitives

### DIFF
--- a/src/components/atoms/switch/index.tsx
+++ b/src/components/atoms/switch/index.tsx
@@ -1,14 +1,29 @@
 import React from "react"
-import * as Switch from "@radix-ui/react-switch"
+import clsx from "clsx"
+import * as _Switch from "@radix-ui/react-switch"
 
-type P = {
-  checked: boolean
-}
+/**
+ * A controlled switch component atom.
+ */
+function Switch(props: _Switch.SwitchProps) {
+  const { checked } = props
 
-export default function (props: P) {
   return (
-    <Switch.Root>
-      <Switch.Thumb />
-    </Switch.Root>
+    <_Switch.Root
+      {...props}
+      className={clsx("w-8 h-[18px] rounded-full", {
+        "bg-gray-300": !checked,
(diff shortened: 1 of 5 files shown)


src/components/atoms/switch/index.tsx
```diff
@@ -0,0 +1,29 @@
+import React from "react"
+import clsx from "clsx"
+import * as _Switch from "@radix-ui/react-switch"
```
pKorsholm, line 3: Nit: Rename this to `RadixSwitch` as we have done in other components that Radix components

src/components/atoms/switch/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * A controlled switch component atom.
+ */
+function Switch(props: _Switch.SwitchProps) {
+  const { checked } = props
+
+  return (
+    <_Switch.Root
+      {...props}
+      className={clsx("w-8 h-[18px] rounded-full", {
```
pKorsholm, line 14: We could also add a transition to the background color as well

We could consider adding this to the `tailwind.config.ts` as well in the `transitionProperty`s. I'm thinking we could add a `transition-bg` class for background color animation: 
```javascript
 transitionProperty: {
        ...
        bg: "background-color",
      },
```
 wdyt?

src/components/atoms/switch/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+        "bg-gray-300": !checked,
+        "bg-violet-60": checked,
+      })}
+    >
+      <_Switch.Thumb
+        className={clsx("w-2 h-2 bg-white rounded-full block transition-all", {
```
pKorsholm, line 20: nit: we could limit the transition class to just translations here

pKorsholm (commented):
Love the animation of the "thumb", very nice.

I just have a couple of comments 😄

srindom (approved):
LGTM!

PR #453, opened 2022-03-25
fix: replace the endpoint for fetching customers of a group

fPolic:
**What**
- ~~use `/admin/customers` in combination with the `groups` param to fetch customers for the groups details page instead of using `/admin/customer-groups/:id/customers`~~ 
- remove customer groups filter on the details page
- set metadata filed to `null` if deleted

**Why**
- ~~`/admin/customer-groups/:id/customers` enforces a single `groups` id as a param by design but we have a use case on the CG customers table where we need customers that belong in 2 groups~~

### DIFF
--- a/src/domain/customers/groups/customer-group-modal.tsx
+++ b/src/domain/customers/groups/customer-group-modal.tsx
@@ -36,7 +36,16 @@ function CustomerGroupModal(props: CustomerGroupModalProps) {
 
   const onSubmit = (data) => {
     const meta = {}
+    const initial = props.initialData?.metadata || {}
+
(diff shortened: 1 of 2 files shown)


srindom (commented):
I actually think that the issue here is that the design is not entirely correct when it comes to the filtering capabilities. It shouldn't be possible to go to the customer group customer overview and then filter based on customers in other groups. Instead, we should leave it without the filtering capabilities for now and then revisit it when we add more advanced filtering to the Customer table overview. 

Sorry for the confusion. 

Would be nice if you could update this to remove the "Groups" filter and just leave it with the raw customer overview for now.

srindom (approved):
LGTM!

PR #455, opened 2022-03-28
Fix/accordion animation

kasperkristensen:
**What**

Attempt at fixing accordion animation. Still not perfect, but this is the best solution I can come up with while maintaining `forceMount` on `Item.Content`.
It ensures that the open animation plays for all items, but it messes a bit with the close animation, still think it is acceptable though. See attached video:

https://user-images.githubusercontent.com/45367945/160386283-f2d06dd9-b073-46b1-830f-4ff014d26be5.mov

**Note** 

The open animation does not work in firefox and safari, see: https://github.com/radix-ui/primitives/issues/1265.

### DIFF
--- a/src/components/organisms/accordion/index.tsx
+++ b/src/components/organisms/accordion/index.tsx
@@ -1,4 +1,5 @@
 import * as AccordionPrimitive from "@radix-ui/react-accordion"
+import clsx from "clsx"
 import React from "react"
 import Button from "../../fundamentals/button"
 import InfoTooltip from "../../molecules/info-tooltip"
@@ -8,6 +9,7 @@ type AccordionItemProps = AccordionPrimitive.AccordionItemProps & {
   description?: string
   required?: boolean
   tooltip?: string
+  forceMountContent: true | undefined
 }
 
 const Accordion: React.FC<
@@ -29,6 +31,7 @@ const Item: React.FC<AccordionItemProps> = ({
   required,
   tooltip,
   children,
+  forceMountContent = undefined,
   ...props
 }) => {
   return (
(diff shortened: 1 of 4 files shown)


src/components/organisms/accordion/index.tsx
```diff
@@ -123,12 +48,17 @@ const Item: React.FC<AccordionItemProps> = ({
...
+        <AccordionPrimitive.Content
+          forceMount
+          className="overflow-hidden radix-state-open:animate-accordion-open radix-state-closed:animate-accordion-close"
+        >
+          <div className="text-grey-50 inter-base-regular">
+            {description && <p>{description}</p>}
+            <div className="mt-large">{children}</div>
+          </div>
+        </AccordionPrimitive.Content>
+      </div>
```
pKorsholm, line 61: I think this can also be solved in the following way, by adding the close animation to the content of the `<AccordionPrimitive.Content>`.

I'm not 100% that this covers all edge cases but i couldn't break it

```suggestion
      <AccordionPrimitive.Content
        forceMount
        className={clsx(
          "overflow-hidden radix-state-open:animate-accordion-open radix-state-closed:animate-accordion-close"
        )}
      >
        <div className="text-grey-50 inter-base-regular group-radix-state-closed:animate-accordion-close">
          {description && <p>{description}</p>}
          <div className="mt-large">{children}</div>
        </div>
      </AccordionPrimitive.Content>
```

pKorsholm (commented):
looks good generally 😄  I have one suggestion which might make it flow a bit more smoothly

https://user-images.githubusercontent.com/88927411/160571820-48081fa0-30d8-416d-8b76-cd144816a961.mov

y

src/components/organisms/accordion/index.tsx
kasperkristensen: Works like a charm! Also made forceMount optional as we might not always need it 😄

zakariaelas (approved):
Very nice @kasperkristensen! LGTM 👍

PR #470, opened 2022-03-28
feat: table actions toaster

zakariaelas:
### What

- adds a declarative `Toaster` component
- adds a `HotKeyAction` component
- adds a `TableToasterContainer` component

### Why
- `react-hot-toast` provides an imperative API for dealing with toasters. Although this is very useful and nice for a lot of use cases (i.e: onError/onSuccess callbacks), sometimes, it might be better to have a declarative approach. For example, when handling table "batch" actions (i.e: select multiple rows then delete), `react-table` will return a `selectedFlatRows` field which will hold all the rows which are currently selected. We can use this piece of state to determine whether we should render a toaster or not. Instead of making a `useEffect` and subscribing to some piece of state everytime we want to use it this way, I have abstracted this functionality in a `<Toaster visible={your_state} />` component

- The `HotKeyAction` component is a simple component which is responsible for rendering a label (to describe what the hot key does) and an icon. It also take a `hotKey` and `onAction` props which will be used to map the given `onAction` callback to the given hotkey. 

- The `TableToasterContainer` component is the toaster container which will wrap any children components.

### DIFF
--- a/src/components/fundamentals/icons/backspace-icon/index.tsx
+++ b/src/components/fundamentals/icons/backspace-icon/index.tsx
@@ -1,0 +1,42 @@
+import React from "react"
+import IconProps from "../types/icon-type"
+
+const BackspaceIcon: React.FC<IconProps> = ({
+  size = "16",
+  color = "currentColor",
+  ...attributes
+}) => {
+  return (
+    <svg
+      width={size}
+      height={size}
+      viewBox="0 0 16 16"
+      fill="none"
+      xmlns="http://www.w3.org/2000/svg"
+    >
+      <path
+        d="M7.5 6.66675L10.1667 9.33341"
+        stroke={color}
+        stroke-width="1.5"
+        stroke-linecap="round"
+        stroke-linejoin="round"
+      />
+      <path
+        d="M10.1667 6.66675L7.5 9.33341"
+        stroke={color}
+        stroke-width="1.5"
+        stroke-linecap="round"
+        stroke-linejoin="round"
+      />
+      <path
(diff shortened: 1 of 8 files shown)


src/components/fundamentals/icons/backspace-icon/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+      width={size}
+      height={size}
+      viewBox="0 0 16 16"
+      fill="none"
+      xmlns="http://www.w3.org/2000/svg"
+    >
+      <path
+        d="M7.5 6.66675L10.1667 9.33341"
+        stroke={color}
+        stroke-width="1.5"
```
kasperkristensen, line 20: nit: update all kebab-cased props too camelCase

src/components/molecules/table-toaster/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+      {...toast?.ariaProps}
+    >
+      <div className="flex items-center rounded-rounded bg-grey-90 px-base py-3.5">
+        {children}
+      </div>
+    </div>
+  )
+}
+
+export const HotKeyAction = ({ label, hotKey, icon, onAction }) => {
```
kasperkristensen, line 30: Should this be removed? 😄

kasperkristensen (approved):
Great work Zak LGMT just have a single nit regarding svg properties, and guess we could remove the HotKeyAction component from table toaster as it's present in HotKeyAction.tsx 😄

fPolic (approved):
Nice 👍

PR #472, opened 2022-03-29
feat: extend tooltip component

kasperkristensen:
**What**
- Changes `InfoTooltip` to `IconTooltip`
- The default type is `info`, and as such no changes are needed to the existing usage of the component.
- Adds types to component
- has 3 `type`'s: `info`, `warning` and `error`


pKorsholm (approved):
lgtm

PR #473, opened 2022-03-29
fix: modal showing unintentionally in prod and housekeeping

kasperkristensen:
**What**

- Prevents Add Shipping Option modal from showing when changing between regions in prod (tested by building and serving site)
- General housekeeping of the region domain:
  - fixed some ts errors
  - updated the look of the cancel buttons in modals to reflect latest changes to design in Figma


olivermrbl (approved):
LGTM!

PR #475, opened 2022-03-30
Feat: Refactor promotions table

pKorsholm:
**What**
- Rename discount -> promotions in sidebar
- Rename discount index header -> promotions
- Rename files with discount in filename to have promotion instead
- Refactor promotions table to match figma design better
  - Add row actions 
  - Add products column
  - Adjust color on statusdot
  - Rename "usage count" to "Redemptions"
    - truncate values at every 10^3 suffixing the correct unit (k, m, b) (if you have more than 1000 billion usage count it is not truncated)

**How** 
- Add `use-copy-promotion` hook for promotion duplication
- Add `use-promotion-row-actions` hook for row actions
- Adjust `use-promotion-columns` (previously `use-discount-columns`) to include updated styling

image


src/components/templates/discount-table/use-promotion-columns.tsx
```diff
@@ -0,0 +1,223 @@
...
+    ],
+    []
+  )
+
+  return [columns]
+}
+
+const getUsageCount = (usageCount: number) => {
+  switch (true) {
+    case usageCount > 9999999:
```
pKorsholm, line 182: A different solution could be something like this: 

I prefer the switch because if the immediate readability

```
  const factor = Math.floor((`${usageCount}`.length - 1) / 3)
  const units = ["", "k", "m", "b"]
  return `${Math.floor(usageCount / Math.pow(10, 3 * factor))}${units[factor]}`
```

olivermrbl (changes requested):
So nice @pKorsholm !! Have added a couple of comments and requests 👍

src/components/templates/discount-table/use-promotion-columns.tsx
pKorsholm: yeah, I agree, highly unlikely, though I thought it better to just cover our bases 😄

src/components/templates/discount-table/use-promotion-columns.tsx
```diff
@@ -0,0 +1,223 @@
...
+  switch (condition.type) {
+    case "product_types":
+      return condition.product_types.length
+    case "products":
+      return condition.products.length
+    case "product_collections":
+      return condition.product_collections.length
+    case "product_tags":
+      return condition.product_tags.length
+    case "custosmer_group":
```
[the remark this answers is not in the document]
olivermrbl: Sorry for this one. Should be plural 😢

src/components/templates/discount-table/use-promotion-columns.tsx
```diff
@@ -0,0 +1,223 @@
...
+  switch (condition.type) {
+    case "product_types":
+      return condition.product_types.length
+    case "products":
+      return condition.products.length
+    case "product_collections":
+      return condition.product_collections.length
+    case "product_tags":
+      return condition.product_tags.length
+    case "customer_group":
```
olivermrbl, line 201:
```suggestion
    case "customer_groups":
```

olivermrbl: My bad first time around 😢

PR #480, opened 2022-03-31
Feat: focus modal for promotion creation and editing

pKorsholm:
**What**
- Add focus modal for Promotion creation and editing
- Add initial promotions detail page

**How**
- update accordion component
  - Add support for different levels of heading sizes with different levels of padding
  - Add support for custom trigger components
  - Change heigh animation to use max- and min-height (animating from 0, 0 to none, radix content height), this enables accordion sections to grow and shrink dynamically.
-  update `discount-form-context-ts`
  - Add `hasStartDate` to toggle start date 
  - Add `handleConfigurationChanged` to control the Configuration accordian item
  - Remove notification provider along with save methods
- update mapper component
  - explicitly only include `valid_duration` for dynamic discounts
- discount form index 
  - wrap the discount from in a focus modal and include
- discount form sections
  - general: reformat general section and change editable fields depending on `isEdit` and discount type
  - promotion types: add radiogroup to change promotion type when creating a discount (otherwise it's disabled)
  - configuration: Add sub accordion with manual control due to the radix classes causing errors in nested accordions. Use Promotion form to handle selecting and deselecting configurations.
- discount index: open modal instead of redirecting to new 
- discount new: a page with an open form
- promotions detail page: 
  - Add initial details page bodycard w. action buttons

*Files to review (due to this being based on detect changes for promotions)*
- `src/components/organisms/accordion/index.tsx`
- `src/domain/discounts/details/index.tsx`
- `src/domain/discounts/discount-form/form/discount-form-context.tsx`
- `src/domain/discounts/discount-form/form/mappers.ts`
- `src/domain/discounts/discount-form/index.tsx`
- `src/domain/discounts/discount-form/sections/promotion-type.tsx`
- `src/domain/discounts/discount-form/sections/conditions.tsx`
- `src/domain/discounts/discount-form/sections/general.tsx`
- `src/domain/discounts/discount-form/sections/configuration.tsx`
- `src/domain/discounts/index.tsx`
- `src/domain/discounts/new/index.tsx`


src/domain/discounts/discount-form/index.tsx
```diff
@@ -1,25 +1,158 @@
...
+    {
+      ...discount,
+      ...(hasStartDate ? { starts_at: startsAt } : {}),
+      ends_at: hasExpiryDate ? endsAt : null,
+    }
+  )
+
+  const submitGhost = async (data) => {
+    if (!isEdit) {
+      console.log(data)
```
kasperkristensen, line 41: rm

src/domain/discounts/discount-form/index.tsx
```diff
@@ -1,25 +1,158 @@
...
+      if (isEdit) {
+        await onUpdate(data)
+      } else {
+        await onSaveAsActive(data)
+      }
+      if (closeForm) {
+        closeForm()
+      }
+    } catch (error) {
+      console.log(error)
```
kasperkristensen, line 69: rm

kasperkristensen (commented):
Great job Phil, almost LGTM 😄 

Some small things to consider/change:

- Some out commented code that should be removed before we merge. 
- I like the addition of `customTrigger` to the Accordion, and reusing it for options you can switch on/off, but we should perhaps consider changing the Accordion so it's only the actual trigger that can change the open/closed state, instead of the entire header, as it might be a bit frustrating that you can easily deselect something if you miss click. 
- The `General` item, always flash open/closes when the page is mounted, I can see you've used a timeout to prevent this for the other Accordion items, so perhaps we should also do it for General.
- I get an error when I try to create a promotion of type `free_shipping` because of line 63 in the mapper: `parseInt((values.rule.value! as unknown) as string, 10),` if the type is `free_shipping` then `values.rule` is undefined.
- Also we should add `navigate(-1)` on the close button in the top right corner of the focus modal

srindom (commented):
LGTM! 

nit (if-minor): add a bit of padding so that you can scroll a bit further down than just to where the section ends :)

image

srindom (approved):

fPolic (approved):
LGTM! 

nit: it would be good to address the initial content flash that @kasperkristensen mentioned before merge

PR #481, opened 2022-03-31
feat: refactor date picker

zakariaelas:
### What

- used radix's new `Select` component to mimic the native select behavior on macOS
- changed our `DatePicker`'s `CustomHeader` component to use dropdowns for the month + year

### How

- created a `NativeSelect` component which is supposed to be lightweight compared to our current `Select` component. See story for usage and demo


src/components/molecules/native-select/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+      <RadixSelect.SelectTrigger
+        className="flex items-center inter-base-semibold gap-3 px-2"
+        {...triggerProps}
+      >
+        <RadixSelect.Value />
+        <RadixSelect.Icon>
+          <ChevronDownIcon size={16} />
+        </RadixSelect.Icon>
+      </RadixSelect.SelectTrigger>
+      <RadixSelect.Content className="rounded-rounded scrollbar-hide border px-2 py-2 border-grey-20 bg-grey-0 w-full flex shadow-dropdown">
```
pKorsholm, line 32: can we add a height property to the Native select component(the time picker has an existing height of 305px)? 

Otherwise the select can grow to the height of the entire screen:
image

pKorsholm: Also should we consider removing the `RadixSelect.ScrollDownButton` and `ScrollUpButton`. They dont seem to be in the design but I don't know if you've discussed this with Ludvig.

pKorsholm (commented):
Overall it looks good 😄  
Definitely the right call to move the header out of the datepicker component 

I just have a couple of comments

src/components/molecules/native-select/index.tsx
zakariaelas: > can we add a height property to the Native select component(the time picker has an existing height of 305px)?

It's actually a bit tricky unfortunately. I think that the select component is supposed to behave that way. It mimics the behavior of native selects in macOS. Unfortunately, giving a max height to the select causes some positioning issues. There is an interesting discussion about this [here](https://github.com/radix-ui/primitives/issues/1193) where the radix team share their thoughts around this. Also, there might be support for setting a max height somewhere down the line (see this [answer](https://github.com/radix-ui/primitives/issues/1193#issuecomment-1050796688)). 

Personally, I think the behavior is not too bad and very acceptable imo, but I agree that being able to set a max-height would be ideal


> Also should we consider removing the RadixSelect.ScrollDownButton and ScrollUpButton. They dont seem to be in the design but I don't know if you've discussed this with Ludvig.

Totally! I thought it was pretty neat from a UX perspective, but will definitely run it by Ludvig and see what he thinks.

src/components/atoms/date-picker/custom-header.tsx
```diff
@@ -0,0 +1,41 @@
+import moment from "moment"
```
srindom, line 1: question (non-blocking): what would it take to avoid the moment dependency?

src/components/molecules/native-select/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+}) => {
+  return (
+    <RadixSelect.Root {...props}>
+      <RadixSelect.SelectTrigger
+        className="flex items-center inter-base-semibold gap-3 px-2"
+        {...triggerProps}
+      >
+        <RadixSelect.Value />
+        <RadixSelect.Icon>
+          <ChevronDownIcon size={16} />
```
srindom, line 29: nit: maybe create `ICON_SIZE` to avoid magic numbers

srindom (commented):
Nice work - think Phil has some great comments that would be nice to address :)

src/components/molecules/native-select/index.tsx
pKorsholm: In that case we should perhaps to something like +- 20 years 😄

src/components/atoms/date-picker/custom-header.tsx
```diff
@@ -1,34 +1,37 @@
...
 import { ReactDatePickerCustomHeaderProps } from "react-datepicker"
-import { months, years } from "./utils"
+import { monthNames, getYearRange } from "./utils"
 
 const CustomHeader = ({
   date,
   changeYear,
   changeMonth,
 }: ReactDatePickerCustomHeaderProps) => {
+  const month = date.toLocaleString("default", { month: "long" })
```
srindom, line 11: todo: fix the locale so that it won't depend on the system locale. Thinking that if this runs on a computer with e.g. da-DK locale you would get "Juni" and not "June" which would break the implementation. Maybe we could use `getUTCMonth` and make the implementation depend on the numerical value of the month instead of the name?

srindom (commented):
Almost ready - I think using `toLocaleString` will give different output depending on the system settings.

src/components/atoms/date-picker/custom-header.tsx
zakariaelas: let me know if the adjustment makes sense. I just used the `getMonth()` method which returns the numerical month (0-12) and use the `monthNames` map to get the month name in english

src/components/atoms/date-picker/custom-header.tsx
```diff
@@ -8,14 +8,15 @@ const CustomHeader = ({
   changeYear,
   changeMonth,
 }: ReactDatePickerCustomHeaderProps) => {
-  const month = date.toLocaleString("default", { month: "long" })
+  const month = date.getMonth()
+  const monthName = monthNames[month]
```
srindom, line 12: praise: 🙌

srindom (approved):
LGTM!

PR #482, opened 2022-03-31
feat: change styling for single select

zakariaelas:
### What

- use radio as a start adornment when `Select` is single select

### How

- changed the `Option` component to use different adornments depending on the value of the `isMulti` prop


pKorsholm (approved):
lgtm 😄

PR #483, opened 2022-03-31
feat: detect dirty state

kasperkristensen:
**What**

Implements a new approach to handling the detection of dirty state in forms and then displaying an actionable notification. Only implemented in Product domain in this PR, will also update Gift Cards after this is merged.

The previous approach as implemented in #362, had some issues as it had to work for forms with various degrees of state controlled by `useForm`, `useState` and other forms of state management, various amounts of possible actions, etc. 

**How**

This approach opts for providing some fundamentals and utility function that can be used to implement the pattern with adaptations required for each form.

The approach is a bit more verbose, as opposed to simply passing props to a hook, but avoids the issues of the previous iteration. 

**Also in this PR**

- Fixes to `ProductForm`
  - Makes it possible to change thumbnails on products.
- Updates design of regular notifications so they are in line with the new design.


src/domain/products/product-form/form/product-form-context.tsx
```diff
@@ -45,20 +65,25 @@ export const ProductFormProvider = ({
...
-
   const methods = useForm()
 
   React.useEffect(() => {
+    if (JSON.stringify(product.images) !== JSON.stringify(images)) {
+      setHasImagesChanged(true)
+    } else {
+      setHasImagesChanged(false)
+    }
+  }, [JSON.stringify(images)])
```
fPolic, line 84: suggestion: I think we could handle `images` change detection in a less "expensive" way without using `JSON.stringify()`. Wdyt about removing this `useEffect` and adding `setHasImagesChanged(true)` call to `appendImage/removeImage` and also setting the flag to `false` in the `resetForm`.

src/domain/products/edit.tsx
```diff
@@ -21,8 +25,10 @@ const EditProductPage = ({ id }) => {
   const notification = useNotification()
   const { product, isLoading } = useAdminProduct(id)
   const updateProduct = useAdminUpdateProduct(id)
+  const [submitting, setSubmitting] = useState(false)
 
   const onSubmit = async (data) => {
+    setSubmitting(true)
```
fPolic, line 31: this flag also needs to be set to `false` if an image upload fails for some reason (L:40) otherwise the form will be "stuck" in the submitting state.

src/domain/products/edit.tsx
```diff
@@ -60,39 +68,64 @@ const EditProductPage = ({ id }) => {
...
+    formState.dirtyFields,
+    additionalDirtyState
+  )
+
+  useEffect(() => {
+    if (isDirty) {
+      setVisible(true)
+    } else {
+      setVisible(false)
+    }
```
fPolic, line 102: nit:
```suggestion
      setVisible(isDirty)
```

fPolic (commented):
Nice work 🚀  UX is really good!
LGTM if the image error case is covered.

src/domain/products/edit.tsx
```diff
@@ -60,39 +68,64 @@ const EditProductPage = ({ id }) => {
       onSubmit={onSubmit}
     >
       <ProductForm product={product} isEdit />
```
fPolic, line 70: Also, I noticed a flash of old values on submit. A hacky fix for this could be that we temporarily, while submitting, send form data to `ProductForm` instead of a product from the hook. e.g.
`<ProductForm product={submitting ? formValuesToUpdateProductMapper(formData) : product} isEdit />`

src/domain/products/product-form/form/product-form-context.tsx
kasperkristensen: Agree that it would be nice to use a less expensive approach, but the issue with relying on append/remove is that the design allows you to reorganise image order, which would not be detected by calling these methods. But at the same time we don't actually support ordering images in the backend, so will have to check in if it's something that we are going to support at some point :)

src/domain/products/edit.tsx
kasperkristensen: I've also noticed it, and actually it's a recurring issue with all of our forms. It's just less noticeable with the old design as you would scroll all the way down to the page to save and not see most of the fields

src/domain/products/product-form/form/product-form-context.tsx
olivermrbl: Think it's a good suggestion @fPolic. We will most likely build out image ordering on products as part of the Extend Product API epic, that we have coming up, but until then I think we should remove both the comparison here and the ability to drag images. Let me know, if this works @kasperkristensen. If so, I'll create a ticket for the latter

src/domain/products/product-form/sections/images.tsx
```diff
@@ -54,37 +55,38 @@ const columns = [
...
-            setValue("thumbnail", value)
+        <Controller
+          name="thumbnail"
+          control={control}
+          render={({ value, onChange }) => {
+            return (
+              <RadioGroup.Root
+                value={value}
+                onValueChange={(value) => {
+                  console.log(value)
```
olivermrbl, line 76: rm

olivermrbl (approved):
LGTM with comments! Great work @kasperkristensen 👍

src/domain/products/product-form/form/product-form-context.tsx
kasperkristensen: @olivermrbl Definetly makes sense to remove the re-ordering as long as it is not supported. I'll remove the comparison in this PR, and then we will open a seperate PR to remove the dragable table? :)

olivermrbl: Yes, sounds good!

PR #484, opened 2022-03-31
fix: radio group styling

zakariaelas:
### What

- add a border to radio item when selected

### How

- Given that the label element is actually a parent of the radio component, there is no css selector for selecting the parent when the radio is checked. The solution adds an extra `RadioPrimitive.Indicator` which is purely stylistic. It simply renders a violet outline on the parent. The `RadioPrimitive.Indicator` is a component which radix renders when the radio item is checked. You can think about having two instances of this component as a way of saying "when the radio is selected, indicate this by rendering an outline + a little radio icon"

would love to hear if any of you guys can come up with a cleaner and more elegant solution!

### DIFF
--- a/src/assets/styles/global.css
+++ b/src/assets/styles/global.css
@@ -160,7 +160,7 @@
     @apply font-mono text-xsmall leading-xsmall font-bold;
   }
 
-  .radio-outer-ring > span[data-state="checked"] {
+  .radio-outer-ring > span.indicator[data-state="checked"] {
     @apply rounded-circle shadow-[#7C3AED] shadow-[0_0_0_2px];
   }
 }
--- a/src/assets/styles/global.css
+++ b/src/assets/styles/global.css
@@ -161,7 +161,7 @@
   }
 
   .radio-outer-ring > span.indicator[data-state="checked"] {
-    @apply rounded-circle shadow-[#7C3AED] shadow-[0_0_0_2px];
+    @apply rounded-circle shadow-violet-60 shadow-[0_0_0_2px];
   }
 }
 
--- a/src/components/organisms/radio-group/index.tsx
+++ b/src/components/organisms/radio-group/index.tsx
@@ -27,7 +27,8 @@ const Item = ({
   return (
     <label
       className={clsx(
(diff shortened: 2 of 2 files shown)


src/assets/styles/global.css
```diff
@@ -160,7 +160,7 @@
     @apply font-mono text-xsmall leading-xsmall font-bold;
   }
 
-  .radio-outer-ring > span[data-state="checked"] {
+  .radio-outer-ring > span.indicator[data-state="checked"] {
     @apply rounded-circle shadow-[#7C3AED] shadow-[0_0_0_2px];
```
pKorsholm, line 164:
```suggestion
    @apply rounded-circle shadow-violet-60 shadow-[0_0_0_2px];
```

src/components/organisms/radio-group/index.tsx
```diff
@@ -41,10 +42,16 @@ const Item = ({
...
           className={clsx(
-            "flex items-center justify-center w-full h-full relative",
+            "indicator flex items-center justify-center w-full h-full relative",
             "after:absolute after:inset-0 after:m-auto after:block after:w-[12px] after:h-[12px] after:bg-violet-60 after:rounded-circle"
           )}
         />
+        <RadioGroupPrimitive.Indicator
+          aria-hidden="true"
+          className={clsx(
+            "absolute inset-0 shadow-[#7C3AED] shadow-[0_0_0_2px] rounded-base"
```
pKorsholm, line 52:
```suggestion
            "absolute inset-0 shadow-violet-60 shadow-[0_0_0_2px] rounded-base"
```

pKorsholm (commented):
I spent a bit of time looking at possible different solutions, though I didn't find any nice ones. I think this is okay, but I think we should drop a comment regarding the two indicators.

kasperkristensen:
Also looks good to me, and I think it's perfectly fine as is. One other option I think could be viable is to use a [MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver/observe), and listen for changes to the `[data-checked]` attribute which changes when a Radio is selected/unselected, and then use that to update the styling of the containing div.

srindom (approved):

PR #487, opened 2022-04-04
Feat/promotion settings

pKorsholm:
**What** 
- Add the "settings" bodycard to Promotion details

**How**
- using a hook to get promotion settings
- add a disabled property to actionables (default is false to change nothing else) 
- add a `NumberedItem` component for each setting


src/domain/discounts/details/use-promotion-settings.tsx
```diff
@@ -0,0 +1,333 @@
...
+        icon: <EditIcon size={20} />,
+        onClick: () => openWithItems(["general"]),
+      },
+      {
+        label: "Delete setting",
+        icon: <TrashIcon size={20} />,
+        variant: "danger",
+        disabled: true,
+        onClick: () => {},
+      },
```
olivermrbl, line 126: nit(non-blocking): Would it make sense to omit "Delete setting" if it's unavailable instead of showing it as disabled.

src/domain/discounts/details/use-promotion-settings.tsx
```diff
@@ -0,0 +1,333 @@
...
+        }
+      })
+      .filter((s) => s !== null)
+    conditions.push(...(displaySettings as displaySetting[]))
+  }
+
+  return conditions
+}
+
+removeNullish({})
```
olivermrbl, line 331: rm

src/domain/discounts/details/use-promotion-settings.tsx
```diff
@@ -0,0 +1,333 @@
...
+          icon: <TrashIcon size={20} />,
+          variant: "danger",
+          onClick: async () =>
+            await updateDiscount.mutateAsync(
+              { usage_limit: null },
+              {
+                onSuccess: () => {
+                  notification(
+                    "Success",
+                    "Discount end date removed",
```
olivermrbl, line 206:
```suggestion
                    "Discount usage limit removed",
```

src/domain/discounts/details/use-promotion-settings.tsx
```diff
@@ -0,0 +1,333 @@
...
+          icon: <TrashIcon size={20} />,
+          variant: "danger",
+          onClick: async () =>
+            await updateDiscount.mutateAsync(
+              { valid_duration: null },
+              {
+                onSuccess: () => {
+                  notification(
+                    "Success",
+                    "Discount end date removed",
```
olivermrbl, line 246:
```suggestion
                    "Discount duration setting removed",
```

src/domain/discounts/details/use-promotion-settings.tsx
```diff
@@ -0,0 +1,333 @@
...
+                              }),
+                            })),
+                        ],
+                      },
+                    },
+                    {
+                      onSuccess: () => {
+                        notification(
+                          "Success",
+                          "Discount end date removed",
```
olivermrbl, line 310:
```suggestion
                          "Discount conditions removed",
```

src/domain/discounts/details/use-promotion-settings.tsx
```diff
@@ -0,0 +1,333 @@
...
+        icon: <EditIcon size={20} />,
+        onClick: () => openWithItems(["configuration"]),
+      },
+      {
+        label: "Delete setting",
+        icon: <TrashIcon size={20} />,
+        variant: "danger",
+        disabled: true,
+        onClick: () => {},
+      },
```
olivermrbl, line 145: nit: Same as other comment.

olivermrbl (commented):
LGTM! Really great work @pKorsholm 🚀  Works really nice!

pKorsholm:
I've been through the comments and initially have removed the disabled actionables.

src/domain/discounts/details/use-promotion-settings.tsx
pKorsholm: Maybe, I have removed them for now, you can see how you like it. I initially had them because the actionables seemed a bit empty without. However, I think if we add them we should also add a tooltip that says you cannot delete this setting. I think then it would make sense, otherwise we shouldn't have them. 

wdyt?

olivermrbl (approved):
LGTM! 🚀  Great work

PR #488, opened 2022-04-04
fix: gift card domain and omit images on create product if no images are present

kasperkristensen:
**What**

- Adds the ability to upload thumbnail on Gift Card creation
- Changes the `/manage` page to use FormContextProvider pattern from other domains
- Adds uploading and editing images on `/manage`
- Adds ability to add new denominations to a already existing Gift Card
- General clean up in the `GiftCard` domain
- Implements `detect-dirty-state` pattern in `/manage` form.

**Also**
- Omits images from `create product` if no images have been uploaded
- Moves `consolidateImages` to `utilities` folder


srindom (approved):
LGTM

PR #489, opened 2022-04-04
feat: add price list table

zakariaelas:
### What

- adds a price list table
- adds a `useSelectionColumn` hook that we can use whenever we need a column for selecting multiple rows

### How

- For the most part, follows the same approach with tables we've been implementing so far

### DIFF
--- a/src/components/fundamentals/icons/cash-icon/index.tsx
+++ b/src/components/fundamentals/icons/cash-icon/index.tsx
@@ -1,0 +1,99 @@
+import React from "react"
+import IconProps from "../types/icon-type"
+
+const CashIcon: React.FC<IconProps> = ({
+  size = "20",
+  color = "currentColor",
+  ...attributes
+}) => {
+  return (
+    <svg
+      width={size}
+      height={size}
+      viewBox="0 0 20 20"
+      fill="none"
+      xmlns="http://www.w3.org/2000/svg"
+      {...attributes}
+    >
+      <path
+        d="M3.75 7.5H2.5"
+        stroke={color}
+        stroke-width="1.5"
+        stroke-linecap="round"
+        stroke-linejoin="round"
+      />
+      <path
+        d="M3.75 12.5H2.5"
+        stroke={color}
+        stroke-width="1.5"
+        stroke-linecap="round"
+        stroke-linejoin="round"
+      />
+      <path
+        d="M3.33366 10H1.66699"
+        stroke={color}
+        stroke-width="1.5"
+        stroke-linecap="round"
+        stroke-linejoin="round"
+      />
+      <path
+        d="M4.69727 4.6967C5.74616 3.64781 7.08253 2.9335 8.53739 2.64411C9.99225 2.35472 11.5002 2.50325 12.8707 3.07091C14.2411 3.63856 15.4125 4.59986 16.2366 5.83323C17.0607 7.0666 17.5006 8.51664 17.5006 10C17.5006 11.4834 17.0607 12.9334 16.2366 14.1668C15.4125 15.4001 14.2411 16.3614 12.8707 16.9291C11.5002 17.4968 9.99225 17.6453 8.53739 17.3559C7.08253 17.0665 5.74616 16.3522 4.69727 15.3033"
+        stroke={color}
+        stroke-width="1.5"
+        stroke-linecap="round"
+        stroke-linejoin="round"
+      />
+      <path
+        d="M10 6.25V7.00737"
+        stroke={color}
+        stroke-width="1.5"
+        stroke-linecap="round"
+        stroke-linejoin="round"
+      />
+      <path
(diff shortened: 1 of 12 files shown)


src/components/templates/price-list-table/index.tsx
```diff
@@ -0,0 +1,271 @@
...
+          <div className="pr-base">
+            <HotKeyAction
+              label="Delete"
+              hotKey="backspace"
+              icon={<BackspaceIcon />}
+              onAction={() => console.log("clicked")}
+            />
+          </div>
+        </TableToasterContainer>
+      </Toaster>
```
olivermrbl, line 245: nit: Would comment this out until we can make use of it

src/components/templates/price-list-table/index.tsx
```diff
@@ -0,0 +1,271 @@
...
+
+const LoadingContainer = ({ isLoading, children }) => {
+  return isLoading ? (
+    <div className="w-full pt-2xlarge flex items-center justify-center">
+      <Spinner size={"large"} variant={"secondary"} />
+    </div>
+  ) : (
+    children
+  )
+}
```
olivermrbl, line 258: nit: Think this component is quite nifty. Should we refactor it to be its own organism / molecule?

src/components/templates/price-list-table/use-price-list-filters.tsx
```diff
@@ -0,0 +1,505 @@
...
+  if (as.size !== bs.size) {
+    return false
+  }
+  for (const a of as) {
+    if (!bs.has(a)) {
+      return false
+    }
+  }
+  return true
+}
```
olivermrbl, line 108: nit: Move to a util function

src/hooks/use-selection-column.tsx
```diff
@@ -0,0 +1,39 @@
...
+          label=""
+          ref={resolvedRef}
+          {...rest}
+        />
+      </div>
+    )
+  }
+)
+
+export const useSelectionColumn = () => {
```
olivermrbl, line 27: praise: I like it! 😍

olivermrbl (commented):
Very great work @zakariaelas ! I've added a couple of comments here and there, that I would like your input  🙌

olivermrbl:
> I can think of few alternatives here:
> 
> 1. **The duplicate operation should not duplicate the `prices` array**: This one makes sense imo, but perhaps users might get a bit disappointed by the prices list not being copied over as entering the prices is probably the most cumbersome task. If we end up implementing this, I suggest adding a short message to explain what's being copied.
> 2. **Remove the duplicate action from the table and support it in the detail view instead**: as we'll most likely be fetching the prices array in the detail view, we can leverage that to duplicate the price list.
> 3. **Ignore the performance concern**: keep things as they are right now (fetch the `prices` array with every price list) until we run into performance issues.
> 
> I personally lean towards option 2. Let me know what you guys think. Might be overthinking a little at this stage, but thought it'd be nice to hear your thoughts on this 😄

I agree! Think 2. is the best option to go with here now. The first one is nice, but without the prices, it kinda loses its purpose.

src/components/fundamentals/icons/cash-icon/index.tsx
```diff
@@ -0,0 +1,99 @@
+import React from "react"
+import IconProps from "../types/icon-type"
+
+const CashIcon: React.FC<IconProps> = ({
```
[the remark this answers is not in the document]
zakariaelas: Quoting the react [docs](https://reactjs.org/docs/dom-elements.html):

> In React, all DOM properties and attributes (including event handlers) should be camelCased. For example, the HTML attribute tabindex corresponds to the attribute tabIndex in React. The exception is aria-* and data-* attributes, which should be lowercased. For example, you can keep aria-label as aria-label.

When we don't convert the svg attributes to camelCase, we get this error:

image

srindom:
@zakariaelas what's the status on this one?

src/components/templates/price-list-table/index.tsx
```diff
@@ -0,0 +1,271 @@
...
+      initialState: {
+        pageSize: lim,
+        pageIndex: offs / lim,
+      },
+      pageCount: numPages,
+      autoResetPage: false,
+    },
+    usePagination,
+    useRowSelect
+  )
```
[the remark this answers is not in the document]
zakariaelas: Just got to play around @fPolic's work yesterday and I reaaaaally liked his approach. I tried to followed the same approach with the comments and everything 😆

zakariaelas:
@srindom just pushed the latest changes 😄

olivermrbl:
> @srindom just pushed the latest changes 😄

Is the `react-table-config.d.ts` required for us to use the types?

olivermrbl (approved):
LGTM! Also love the comments. Really adds some structure to the files improving overall readability 😍

PR #490, opened 2022-04-04
feat: generate promotion code

kasperkristensen:
**What**

- Adds new component `PromotionCodeInput`
- If the input has no value it will show a button in the top right corner that can be used to generate a promotions code. If the input has a value it will instead show a refresh button that generates a new code. 
- Adds a new utility to generate a random code.

I was considering if we should check if a code already exists so we don't generate codes that already exist, but given how small the chances of this is I decided that it's not worth the request. Let me know if any of you disagree. 

Also would like some input on what you think the format of the generated codes should be? Currently it generates a code in the format `[A-Z0-9]-[A-Z0-9]-[A-Z0-9]-[A-Z0-9]`.


adrien2p:
From my point on view

- the discount code should be generated by the server
- the server when generate the discount code should check if it already exists

srindom (approved):
LGTM.

Wrt. @adrien2p's comment I think this is a very good idea, but maybe something to include as an addition for the future.

zakariaelas (approved):
LGTM! Great work Kasper 👍 !

adrien2p:
Nice @kasperkristensen , @srindom should we create an issue for the addition in order to not forget it 😅

PR #491, opened 2022-04-04
feat: "Add conditions" modal

fPolic:
**What**
- implemented a modal for selecting discount condition type


src/domain/discounts/discount-form/form/discount-form-context.tsx
```diff
@@ -0,0 +1,288 @@
...
+    name: "allocation",
+    value: discount?.allocation || "total",
+  })
+
+  const setAllocation = (value) => {
+    methods.setValue("allocation", value)
+    setPrevAllocation(value)
+  }
+
+  const setConditionType = (value) => methods.setValue("condition_type", value)
```
pKorsholm, line 59: For the discount form we should add some way to manage multiple conditions and condition types in the hook. I’m a bit on the fence regarding how, but I would love to hear your thoughts. 

We could create individual properties in the form for each condition entity (products, types, tags, etc.). This would make it easy to control from a modal perspective in terms of the tables and adding/removing condition entities. However, it would bloat the form quite a bit since the fields would need to be controlled in some way. When we then submit the form we would map the array to their id’s and send them along to the api. 

A different option that is more closely related to how the api accepts the discount conditions on create and update would be to store an array of conditions. This would require us to loop the array (which has a max length of 5) when adding or removing condition entities from the discount, however when committing the changes to the api the conditions would require less mapping and wrangling in order to fit the api spec. This solution also bloats the form less since it’s a single field and not 5.

wdyt?

src/components/organisms/radio-group/index.tsx
```diff
@@ -100,11 +107,14 @@ const SimpleItem: React.FC<RadioGroupSimpleItemProps> = ({
           )}
         />
       </RadioGroupPrimitive.Item>
-      {label && (
+      <div className="ml-small inter-base-regular cursor-pointer w-full">
```
pKorsholm, line 110: Nit: Can we add the `indicator` class to the `RadioGroupPrimitive.Indicator` element. 

This will just add a violet-60 outline to the radio button as well as the dot in the center.

I cant place this comment on the right line but it's on line 105

pKorsholm (commented):
Generally looks great 😄  I think we should add some state management for conditions as well in the form in order to have that in place so we can work concurrently on the different tables. 

What do you think?

src/domain/discounts/discount-form/form/discount-form-context.tsx
zakariaelas: I think I lean more towards an implementation of option 2. To give a bit more detail, I think that we should have a `conditions` array or object in the discount form, then leave each individual condition modal (i.e: product conditions modal, customer groups conditions modal, etc.) the freedom to implement/manage its own form. In other words, each conditions modal will be responsible of getting all the selected choices from the table or whatever, then it should call some method like `addCondition` that is exposed from the `useDiscountForm` hook to "register" the user input conditions.

Does this align with what you had in mind @pKorsholm 😄 ?

zakariaelas (commented):
Really cool! LGTM 🚀

Would love to hear your thoughts on Philip's comments 👍

src/domain/discounts/discount-form/form/discount-form-context.tsx
fPolic: So to summarise (and I agree with you guys) the idea we are going for is to have multiple (sub)forms i.e. we would have reference to data in the root form but delegate logic as much as possible to a nested modal/form.

pKorsholm: Yeah, the individual tables should definitely be responsible for managing their own conditions and adding/removing the entities correctly. @kasperkristensen brought up a good point that if we want to re-render the form a change in a child of the array will not trigger a re-render, then we would need to add an appropriate handler. This might become an issue when we get to showing the list of conditions that apply to a promotion in the modal before we save and update the promotion and thus invalidate the query.

We could consider creating an object for the state where the values map directly to the possible conditions plus a title, i.e.: 
```
{
  products?: {products: {id: string, title: string}[]}

  product_collections?: {product_collections: {id: string, title: string}[]}

  product_types?: {product_types: {id: string, title: string}[]}

  product_tags?: {product_tags: {id: string, title: string}[]}

  customer_groups?: {customer_groups: {id: string, title: string}[]}
}
```

Though this also feels like a hack to get it to work. I think I'm leaning more towards solution 1 and then having the mapper or another utility hook to transform the input for the api.

kasperkristensen (commented):
LGTM 😄 , with some minor things I think needs to be changed:

- Shouldn't the `Add condition` button be moved into the last Accordion item? 
- And not sure if the design has been updated today, but looking at Figma for the initial screen of the modal (where you select which type of condition to add) there seems to be a mismatch between the design and implementation: https://www.figma.com/file/a5Mb7n6lelmmFU28wqQybY/Customer-Groups%2C-Promotions-%26-Pricing?node-id=435%3A15494

fPolic:
@kasperkristensen @pKorsholm @zakariaelas - I think this one is ready for merge since the features we discussed in the comments are out of scope for this PR and are being implemented in other PRs.

src/domain/discounts/discount-form/index.tsx
```diff
@@ -149,6 +161,25 @@ const DiscountForm: React.FC<DiscountFormProps> = ({
...
+                  onClick={() => setShowConditionsModal(true)}
+                  className="mb-[25%] medium:w-7/12 large:w-6/12 small:w-4/5 w-full m-auto mt-4 flex justify-center items-center gap-2
+                    font-semibold
+                    text-small
+                    rounded-xl border border-1 p-2"
+                  role="button"
+                >
+                  <PlusIcon size={18} />
+                  <span>Add Condition</span>
+                </div>
```
pKorsholm, line 175: I think we should make this a button: 
```suggestion
               <Button
                  variant="ghost"
                  size="small"
                  onClick={() => setShowConditionsModal(true)}
                  className="w-full mt-4 flex rounded-rounded border p-2"
                  role="button"
                >
                  <PlusIcon size={20} />
                  <span>Add Condition</span>
                </Button>
```

src/domain/discounts/discount-form/index.tsx
```diff
@@ -106,7 +118,7 @@ const DiscountForm: React.FC<DiscountFormProps> = ({
         </div>
       </FocusModal.Header>
       <FocusModal.Main>
-        <div className="flex justify-center pb-[25%]">
+        <div className="flex justify-center">
```
pKorsholm, line 121:
```suggestion
        <div className="flex justify-center pb-[25%]">
```

pKorsholm (approved):
LGTM 😄 just a few nits as suggestions.

kasperkristensen (approved):
LGTM! There are some issues in the promotion domain, the promotion table does not work because the "Products" colum tries to to display some fields that aren't returned by the endpoint (condtion.products/product_types/etc), and the form mapper tries to map `rule.valid_for` which does not exist any more. But none of it is related to your PR, so I say we merge this and fix the issues in another PR so we can start implementing the different condition screens cc: @pKorsholm @fPolic

PR #494, opened 2022-04-06
[Proposal] Conditions styling

pKorsholm:


src/domain/discounts/discount-form/sections/conditions.tsx
```diff
@@ -1,22 +1,54 @@
...
+  return (
+    <span className="text-grey-50 inter-small-regular">
+      {description}
+      <span className="text-grey-40">
+        {titles.length - 2 > 0 && ` + ${titles.length - 2} more`}
+      </span>
+    </span>
+  )
+}
+
```
pKorsholm, line 44: This needs to be a bit more elegant in terms of how we format the description

src/domain/discounts/discount-form/sections/conditions.tsx
```diff
@@ -1,22 +1,60 @@
...
     </div>
   )
 }
 
+const ConditionSetting = ({ titles }) => {
+  const titleStrings = titles.reduce(
+    (acc, cur) =>
+      acc[1] + cur.length > 38 ? acc : [[...acc[0], cur], acc[1] + cur.length],
+    [[], 0]
+  )
```
kasperkristensen:
```suggestion
  const MAX_TITLES_LENGTH = 38
  
  const titleStrings = titles.reduce(
    (acc, cur) =>
      acc[1] + cur.length > MAX_TITLES_LENGTH ? acc : [[...acc[0], cur], acc[1] + cur.length],
    [[], 0]
  )
```

Maybe add something like this so we avoid magic numbers, and provide a bit of explanation for what the number 38 represents 😄

kasperkristensen (approved):
LGTM

PR #498, opened 2022-04-07
add: generic table for promotion conditions

zakariaelas:
### What

- adds a generic table to be used for different promotion condition modals
- uses same approach as customer-groups + price lists
- supports optional headers
- supports optional filtering
- supports custom rows + headers through the `renderRow` and `renderHeaderGroup` props
- automatically debounces search terms

### DIFF
--- a/src/domain/discounts/discount-form/condition-tables/selectable-table.tsx
+++ b/src/domain/discounts/discount-form/condition-tables/selectable-table.tsx
@@ -5,25 +5,32 @@ import {
   ProductTag,
   ProductType,
 } from "@medusajs/medusa"
-import React, { useEffect, useMemo } from "react"
+import { debounce } from "lodash"
+import React, { useEffect } from "react"
 import {
-  ColumnInstance,
+  Column,
+  HeaderGroup,
+  Row,
   usePagination,
   useRowSelect,
   useTable,
 } from "react-table"
 import Spinner from "../../../../components/atoms/spinner"
 import IndeterminateCheckbox from "../../../../components/molecules/indeterminate-checkbox"
-import Table, { TablePagination } from "../../../../components/molecules/table"
-import { PaginationProps } from "../../../../types/shared"
+import Table, {
+  TablePagination,
+  TableProps,
+} from "../../../../components/molecules/table"
+import useQueryFilters from "../../../../hooks/use-query-filters"
 
 type SelectableTableProps = {
-  showSearch?: boolean
-  objectName?: string
+  resourceName?: string
   label?: string
   isLoading?: boolean
-  pagination: PaginationProps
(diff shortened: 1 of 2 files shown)


src/domain/discounts/discount-form/condition-tables/selectable-table.tsx
```diff
@@ -5,197 +5,183 @@ import {
...
   onChange: (items: string[]) => void
-  onSearch?: (search: string) => void
-}
+  renderRow: (props: { row: Row }) => React.ReactElement
+  renderHeaderGroup?: (props: {
+    headerGroup: HeaderGroup
+  }) => React.ReactElement
+} & ReturnType<typeof useQueryFilters>
 
 export const SelectableTable: React.FC<SelectableTableProps> = ({
```
fPolic, line 49: It would be nice to convert `SelectableTable` to a generic function so we can pass the type to `react-table`

fPolic (approved):
👍

src/domain/discounts/discount-form/condition-tables/selectable-table.tsx
zakariaelas: very true! pushed a fix, lmk what you think 😄

src/domain/discounts/discount-form/condition-tables/selectable-table.tsx
```diff
@@ -1,201 +1,175 @@
...
-}
+  renderRow: (props: { row: Row<T> }) => React.ReactElement
+  renderHeaderGroup?: (props: {
+    headerGroup: HeaderGroup<T>
+  }) => React.ReactElement
+} & ReturnType<typeof useQueryFilters>
 
-export const SelectableTable: React.FC<SelectableTableProps> = ({
-  showSearch = true,
+export const SelectableTable = <T extends object>({
```
fPolic, line 37: I think we could be a bit more precise here and narrow the type: `T = Product | CustomerGroup | ...`

zakariaelas: I think we might be extracting this table down the line so that we can use it across different domains including, but not limited to the ones you suggested. Nevertheless, I'll make the change. Perhaps it's better to narrow right now given that we know for a fact that we'll be using it across those domains 😄 . Lmk what you think 👍

fPolic: It makes sense to me, let's leave it this way now since you already made the change and when we will extract the table to be a template we can change the type.

olivermrbl: Can we merge this? @fPolic @zakariaelas

PR #504, opened 2022-04-08
Feat: Rename discount to promotion

pKorsholm:
**What**
- rename occurances of "discount" to promotion:
  - in the ui
  - in the code where possible (some instances i.e. useAdminCreateDiscount has not been renamed. 
 
We could have some discussion as to what should be renamed in the code. In the future, programming an admin system where the entities are all named the same as in the ui might be easier in terms of communicating in code what it's about.
I think we at least should rename all visual entities as well as the url of discounts to be promotion in the admin in order to align everything on that point.


olivermrbl:
As per our conclusion a couple of weeks ago in regards to keeping "Discount" as our wording shown to users, should we close this one? I'll create a new ticket

PR #511, opened 2022-04-13
feat: PriceList context and new PriceList UI

kasperkristensen:
**What**
- Adds PriceListContextProvider
- Adds form sections
- Adds create new PriceList UI
  - _Note_: I've not added the section for adding prices to the list, as this will be part of a seperate PR.

**Note**
- I assume that we are going to need to modify the `mappers` a bit when we implement functionality to Add/Edit prices, as from the UI we are going to need to group variants by `product_id`, which should not be sent to the backend, as it's not part of the request.


src/domain/pricing/pricing-form/sections/configuration.tsx
```diff
@@ -0,0 +1,199 @@
...
+  }
+  if (config.ends_at) {
+    enabledConfigs.push(ConfigurationField.ENDS_AT)
+  }
+
+  return enabledConfigs
+}
+
+const Configuration: React.FC<ConfigurationProps> = () => {
+  const { customer_groups, isLoading } = useAdminCustomerGroups()
```
zakariaelas, line 37: thought (non-blocking): we probably need to implement some mechanism to handle pagination. Current core implementation only returns 10 customer groups, so we might run into issues sooner than elsewhere (i.e: regions). 

Found this [project](https://www.npmjs.com/package/react-select-async-paginate) that might help us implement this with `react-select`. I personally prefer we track this in a separate ticket and instead focus on getting this PR in 😄 . What do you think?

zakariaelas (approved):
LGTM! Great work @kasperkristensen 👍

I think it might be better to change the base branch to something other than `develop` as there are still some parts that will come from other PRs. I've used the `feat/price-list-details` branch as base in my PR. Might be worth considering it as I'll be reusing most of your work in the details page. What do you think?

src/domain/pricing/pricing-form/sections/configuration.tsx
kasperkristensen: I was also thinking that we need to handle this somehow, the project you linked looks like a great candidate! But I agree we should do that in a separate PR 😄

kasperkristensen:
@zakariaelas Are you fine with me merging it into `feat/price-list-details` or would you prefer something else? 😄

fPolic (approved):
LGTM nice work 👍

zakariaelas:
@kasperkristensen `feat/price-list-details` is perfect!

src/domain/pricing/pricing-form/form-header/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+const FormHeader = (props: PriceListFormProps) => {
+  const { handleSubmit } = usePriceListForm()
+  const notification = useNotification()
+
+  const closeForm = () => {
+    navigate(-1)
+  }
+
+  const createPriceList = useAdminCreatePriceList()
+  const updatePriceList = useAdminUpdatePriceList(props?.id!)
```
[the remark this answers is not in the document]
kasperkristensen: True! Removed the `?`, still need to use a bang notation to tell TS that the `id` always exists, in the cases where `updatePriceList` is used 😄

PR #514, opened 2022-04-13
feat: table fields filter

fPolic:
**What**
- add a component for adding/removing (bulk) table columns to/from the view


src/components/molecules/table-fileds-filter/index.tsx
```diff
@@ -0,0 +1,202 @@
...
+    }
+  }, [contentRef])
+
+  return (
+    <DropdownMenu.Root open={open}>
+      <DropdownMenu.Trigger>
+        <Button
+          onClick={onTriggerClick}
+          variant="secondary"
+          className="rounded rounded-lg h-[32px] px-3 text-small font-semibold text-grey-90"
```
kasperkristensen, line 135:
```suggestion
          className="rounded-lg h-[32px] px-3 text-small font-semibold text-grey-90"
```

src/components/molecules/table-fileds-filter/index.tsx
```diff
@@ -0,0 +1,202 @@
...
+
+/******************** CONTAINERS ********************/
+
+/**
+ * The dropdown menu for selecting currently active table fields.
+ */
+function FieldsMenu(props: FieldsMenuProps) {
+  const { fields, onBlur, selectedFields } = props
+
+  const contentRef = useRef()
```
kasperkristensen, line 87:
```suggestion
  const contentRef = useRef<HTMLDivElement>(null)
```

src/components/molecules/table-fileds-filter/index.tsx
```diff
@@ -0,0 +1,202 @@
...
+
+  return (
+    <div className="flex items-center gap-2">
+      <span className="text-small font-semibold text-gray-500 whitespace-nowrap">
+        Currently editing these fields:
+      </span>
+
+      <div className="flex gap-1 overflow-x-auto no-scrollbar::-webkit-scrollbar">
+        {visibleFields.map((f) => (
+          <Chip key={f.id} {...f} remove={() => removeSelected(f.id)} />
```
kasperkristensen, line 189: nit: Resolve `f` possibly being undefined

src/components/molecules/table-fileds-filter/index.tsx
```diff
@@ -0,0 +1,202 @@
...
+
+  const visibleFields = _selected.map((id) => fields.find((f) => f.id === id))
+
+  return (
+    <div className="flex items-center gap-2">
+      <span className="text-small font-semibold text-gray-500 whitespace-nowrap">
+        Currently editing these fields:
+      </span>
+
+      <div className="flex gap-1 overflow-x-auto no-scrollbar::-webkit-scrollbar">
```
kasperkristensen, line 187:
```suggestion
      <div className="flex gap-1 overflow-x-auto no-scrollbar">
```
Also requires adding the utility classes as mentioned in another comment

src/components/molecules/table-fileds-filter/index.tsx
```diff
@@ -0,0 +1,202 @@
...
+      />
+    </DropdownMenu.Item>
+  )
+}
+
+/******************** CONTAINERS ********************/
+
+/**
+ * The dropdown menu for selecting currently active table fields.
+ */
```
kasperkristensen, line 83: nit: I like adding descriptions to the components, but perhaps we should turn off "valid-jsdoc" in the eslint config in that case.

kasperkristensen (changes requested):
Super clean Frane, really like how you've put this together! I have a couple suggestions, that I think should be addressed, and some tiny nits that I think will make this 100% perfect, but you are free to ignore.

image

As you can see from the picture, it starts to look at little off once the filters start overflowing. I think we could do to things to improve this:

1. The cut off is a bit harsh, and I think we make it a little more pleasing to look at bby adding a slight fade to the end of the div, much like the one that is applied to the header of `BodyCard`s.
2. The scrollbar is pushing the Chips up a bit, I can see that you have added a class of no-scrollbar, but it does not exist so the scroll bar still exists. You could add something like this to `global.css`

```css
@layer utilities {
  // other utilities....

  .no-scrollbar::-webkit-scrollbar {
      display: none;
  }

  .no-scrollbar {
      -ms-overflow-style: none; 
      scrollbar-width: none; 
  }
}
```
But if you opt to hide the scrollbar entirely, then I think it becomes very important to add some other type of visual cue to indicate that it's scrollable (such as adding the above mentioned fade)

src/components/molecules/table-fileds-filter/index.tsx
fPolic: I was experimenting with scrollbar hiding and forgot to remove this class but I'm not sure whether it's a good UX practice if we hide the scrollbar though.

fPolic:
> 1. The cut off is a bit harsh, and I think we make it a little more pleasing to look at bby adding a slight fade to the end of the div, much like the one that is applied to the header of `BodyCard`s.

I agree with you on this one, the component looks a little off with overflow. I was also trying to hide the scrollbar but think that is not the best solution either. I will consult with @ludvig18 regarding the approach here but like your "fade" idea 👍

olivermrbl:
> I will consult with @ludvig18 regarding the approach here but like your "fade" idea

@fPolic Did you get a chance to align with @ludvig18 on the scrollbar yet? 😊

fPolic:
> @fPolic Did you get a chance to align with @ludvig18 on the scrollbar yet? 😊

Yes, changes are implemented in the last commit, we went for a wrap approach instead of "fade".

olivermrbl (commented):
Very nice! Really like the structuring of the components with comments. 

@kasperkristensen Would be great if you could review it once more, when time allows for it 🙌

src/components/molecules/table-fileds-filter/index.tsx
olivermrbl: @fPolic What do you think of this suggestion?

fPolic: I would prefer if we could kill that particular rule - I think we were planning to do the same thing on the core project?

src/components/molecules/table-fileds-filter/index.tsx
```diff
@@ -0,0 +1,200 @@
...
+      {visibleFields.map((f) => (
+        <Chip key={f!.id} remove={() => removeSelected(f!.id)} {...f!} />
+      ))}
+
+      <FieldsMenu
+        fields={sortBy(fields, "id")}
+        onBlur={setSelectedFields}
+        selectedFields={_selected}
+      />
+    </>
```
kasperkristensen, line 196:
```suggestion
    <div className="flex-wrap flex items-center gap-y-2">
      <span className="text-small font-semibold whitespace-nowrap text-gray-500 mr-2">
        Currently editing these fields:
      </span>

      {visibleFields.map((f) => (
        <Chip key={f!.id} remove={() => removeSelected(f!.id)} {...f!} />
      ))}

      <FieldsMenu
        fields={sortBy(fields, "id")}
        onBlur={setSelectedFields}
        selectedFields={_selected}
      />
    </div>
```

nit: adding some vertical spacing between "rows" when wrapped, like in the design.

src/components/molecules/table-fileds-filter/index.tsx
olivermrbl: Yes, I agree!

PR #517, opened 2022-04-15
feat: price input

fPolic:
**What**
- add an organism input component for editing price list item price

### DIFF
--- a/src/components/organisms/price-input/index.tsx
+++ b/src/components/organisms/price-input/index.tsx
@@ -17,14 +17,14 @@ type PriceInputProps = {
  */
 function PriceInput(props: PriceInputProps) {
   const { amount, currency, onAmountChange } = props
-  const { code, symbol, decimal_digits } = currency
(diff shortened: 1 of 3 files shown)


src/components/organisms/price-input/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+            bg-grey-5
+            text-gray-90
+            text-right
+            text-small
+          "
+      />
+
+      <div className="absolute flex items-center h-full top-0 right-3">
+        <span className="text-small text-grey-40 mt-[1px]">{symbol}</span>
+      </div>
```
olivermrbl, line 85: Should use same formatting as CurrencyInput. Right now DKK is shown as `Dkr.`, but should be `kr.`

image

olivermrbl (commented):
Do you think it is possible to add the same "format-as-you-type" functionality as we have in our currency input?

olivermrbl (approved):
LGTM! 😍  Feel free to merge as you please @fPolic

PR #523, opened 2022-04-19
feat: price list details edit prices manually

zakariaelas:
### What

- adds support for editing prices manually in the prices section
- change the `FileInputField` prop to support a react node instead of a simple string. This is useful in cases where we need to pass something more complicated than a simple string
 - changes the collapsible parent and leaf components to support custom actions
 - adds the "add products" modal using a selection table
 - tweaks the price list form price section to include the `ProductPrices` component


src/components/templates/add-products-modal/product-table-config.tsx
```diff
@@ -0,0 +1,92 @@
...
+          ) : (
+            <div className="flex items-center justify-center w-full h-full rounded-soft bg-grey-10">
+              <ImagePlaceholderIcon size={16} />
+            </div>
+          )}
+        </div>
+        <div className="flex flex-col">
+          <span>
+            {original.title}{" "}
+            <span className="text-grey-50">({original.subtitle})</span>
```
kasperkristensen, line 30:
```suggestion
            {original.subtitle && <span className="text-grey-50">({original.subtitle})</span>}
```

nit

src/components/molecules/collapsible-tree/index.tsx
```diff
@@ -101,7 +101,9 @@ const CollapsibleTreeParent: React.FC<CollapsibleTreeParentProps> = ({
         <div className="flex items-center justify-between">
           <div className="gap-x-small flex items-center">{children}</div>
           <div className="flex items-center gap-x-xsmall">
-            <Actionables customTrigger={Trigger()} actions={actions} />
+            {actions && (
```
kasperkristensen, line 104: Guessing that it has nothing to with your PR at all, but seems like the dashed lines between "leafs" are broken, and not extending to the next leaf. Is that also how it looks on your end? 😄

kasperkristensen (approved):
Looks great Zak, just have a tiny nit regarding subtitle in the add products modal 😄

src/components/molecules/collapsible-tree/index.tsx
kasperkristensen: image

This is how it looks on my end 🤨  But again not related to your PR, so feel free to merge 😄

olivermrbl:
Can I get you to merge when nits have been incorporated 👍  @zakariaelas

PR #526, opened 2022-04-20
fix: use default store currency when creating variant price

### LINKED ISSUE #525: bug(products): The price first added to a variant is locked to USD
If I create a product variant on an existing product (or draft), the first price added is always locked to USD, even if USD is not configured as a currency. The problem persists after restart of the Admin panel after currency configuration (see below).

## Prerequisites
Make sure USD isn't configured as the default or store currency. On the Admin panel, go to **Settings** > **Currency**, and configure a non-USD currency (e.g. DKK) as both the **Store currencies** and **Default store currency**.

## Steps to reproduce
On the Admin panel: 
1. Go to **Products** > **New product**
2. Fill in a Name
3. Click **Save as draft**. 
4. Open the new product draft. 
5. On the **Variants** panel, click the three dots and then **Add variant**. 
6. On the **Add variant** modal, click **Add a price**

## Actual behaviour
The added price is locked to USD

## Expected behaviour
The added price to be in DKK (or the configured default currency/any store currency).

----
Medusa v1.2.1
Medusa Admin, 8189ab1c6b5cf2f92a226da30f7d665e7c8b39f9

Mozilla Firefox v99.0.1

olavurdj:
## What
The 'Add variant' dialog now uses the default store currency when adding a price
## Why
Fixes #525 
## How
By passing the store configuration parameters to `usePricesFieldArray`
## Testing
The change has been manually tested against a fresh Medusa 1.2.1 instance configured with Postgres:
- [x] Set the store to only have one currency, and mark that currency as default. Verify that price is created with that currency.
- [x] Add multiple currencies to the store, set one as default. Verify that price is created with the default currency.
- [x] Change the default currency. Verify that subsequent prices are created with the new default (without restart).

## Other
Modified file formatted using ESLint/Prettier with repository provided config

### DIFF
--- a/src/domain/products/details/variants/variant-editor.tsx
+++ b/src/domain/products/details/variants/variant-editor.tsx
@@ -48,11 +48,19 @@ const VariantEditor = ({
     appendPrice,
     deletePrice,
     availableCurrencies,
-  } = usePricesFieldArray(store?.currencies.map((c) => c.code) || [], {
-    control,
-    name: "prices",
-    keyName: "indexId",
-  })
+  } = usePricesFieldArray(
+    store?.currencies.map((c) => c.code) || [],
+    {
+      control,
+      name: "prices",
+      keyName: "indexId",
+    },
+    {
+      defaultAmount: 1000,
+      defaultCurrencyCode:
+        store?.default_currency.code || store?.currencies[0].code || "usd",
+    }
+  )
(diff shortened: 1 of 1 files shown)


kasperkristensen (approved):
LGTM! 😄 Thanks for the contribution

PR #539, opened 2022-04-29
Fix: refund amount on return

pKorsholm:
**What**
- Fix bug where receiving a partial return would refund the amount of entire quantity of products (i.e. receiving a return of 1/2 items would refund 2/2 items) 
- remove additional line for return reason which added padding when a line was selected in a swap or claim

**Why**
- to ensure that refunds are calculated correctly

**How**
- adjust how the total for the return is calculated
- Add a table specifically for receiving returns where reasons cannot be set (since they are set when registering)

### DIFF
--- a/src/components/organisms/rma-select-product-table/index.tsx
+++ b/src/components/organisms/rma-select-product-table/index.tsx
@@ -217,7 +217,7 @@ const RMASelectProductTable: React.FC<RMASelectProductTableProps> = ({
               {checked && !isSwapOrClaim && (
                 <Table.Row className="last:border-b-0 hover:bg-grey-0">
                   <Table.Cell></Table.Cell>
-                  <Table.Cell colspan={2}>
+                  <Table.Cell colSpan={2}>
                     <div className="max-w-[470px] truncate">
                       {toReturn[item.id]?.reason && (
                         <span className="inter-small-regular text-grey-40">
@@ -242,7 +242,7 @@ const RMASelectProductTable: React.FC<RMASelectProductTableProps> = ({
                       )}
                     </div>
                   </Table.Cell>
-                  <Table.Cell colspan={2}>
+                  <Table.Cell colSpan={2}>
                     <div className="flex w-full justify-end mb-small">
                       <Button
                         onClick={() =>
--- a/src/components/organisms/rma-select-product-table/index.tsx
+++ b/src/components/organisms/rma-select-product-table/index.tsx
@@ -214,7 +214,7 @@ const RMASelectProductTable: React.FC<RMASelectProductTableProps> = ({
                   {order.currency_code.toUpperCase()}
                 </Table.Cell>
               </Table.Row>
-              {checked && (
+              {checked && !isSwapOrClaim && (
                 <Table.Row className="last:border-b-0 hover:bg-grey-0">
                   <Table.Cell></Table.Cell>
                   <Table.Cell colspan={2}>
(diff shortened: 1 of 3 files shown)


src/domain/orders/details/returns/receive-menu.tsx
```diff
@@ -36,20 +37,22 @@ const ReceiveMenu: React.FC<ReceiveMenuProps> = ({
   const [quantities, setQuantities] = useState({})
```
kasperkristensen, line 37: since `[quantities, setQuantities]` is not used for anything anymore it should be removed

src/domain/orders/details/returns/receive-menu.tsx
```diff
@@ -7,6 +7,7 @@ import CurrencyInput from "../../../../components/organisms/currency-input"
 import RMASelectProductTable from "../../../../components/organisms/rma-select-product-table"
 import { getErrorMessage } from "../../../../utils/error-messages"
 import { displayAmount } from "../../../../utils/prices"
+import RMASelectReturnProductTable from "../../../../components/organisms/rma-select-receive-product-table"
 
 type ReceiveMenuProps = {
   order: any
```
kasperkristensen, line 13: Since we are updating this I think we should move away from not typing it correctly 😄

kasperkristensen: We can also remove `isSwapOrClaim` since it's not used

src/components/organisms/rma-select-receive-product-table/index.tsx
```diff
@@ -0,0 +1,182 @@
+import clsx from "clsx"
+import React from "react"
+import { formatAmountWithSymbol } from "../../../utils/prices"
+import CheckIcon from "../../fundamentals/icons/check-icon"
+import MinusIcon from "../../fundamentals/icons/minus-icon"
+import PlusIcon from "../../fundamentals/icons/plus-icon"
+import Table from "../../molecules/table"
+
+type RMASelectProductTableProps = {
```
kasperkristensen, line 9: We should type this properly

src/components/organisms/rma-select-receive-product-table/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+                  </div>
+                  <div className="inter-small-regular text-grey-50 flex flex-col ml-4">
+                    <span>
+                      <span className="text-grey-90">{item.title}</span>
+                    </span>
+                    <span>{item?.variant?.title || ""}</span>
+                  </div>
+                </div>
+              </Table.Cell>
+              {/* <Table.Cell className="text-right w-32 pr-8">
```
kasperkristensen, line 137: rm

src/components/organisms/rma-select-receive-product-table/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+    <Table>
+      <Table.HeadRow className="text-grey-50 inter-small-semibold">
+        <Table.HeadCell colSpan={2}>Product Details</Table.HeadCell>
+        <Table.HeadCell className="text-right pr-8">Quantity</Table.HeadCell>
+        <Table.HeadCell className="text-right">Refundable</Table.HeadCell>
+        <Table.HeadCell></Table.HeadCell>
+      </Table.HeadRow>
+      <Table.Body>
+        {allItems.map((item) => {
+          console.log(item)
```
kasperkristensen, line 87: rm

src/components/organisms/rma-select-product-table/index.tsx
```diff
@@ -242,32 +242,30 @@ const RMASelectProductTable: React.FC<RMASelectProductTableProps> = ({
...
-                              ReturnReasonScreen(
-                                pop,
-                                toReturn[item.id]?.reason,
-                                toReturn[item.id]?.note,
-                                customReturnOptions,
-                                imagesOnReturns,
-                                (reason, note, files) =>
-                                  setReturnReason(reason, note, files, item.id)
-                              )
+                  <Table.Cell colspan={2}>
```
kasperkristensen, line 245:
```suggestion
                  <Table.Cell colSpan={2}>
```

src/components/organisms/rma-select-product-table/index.tsx
```diff
@@ -214,7 +214,7 @@ const RMASelectProductTable: React.FC<RMASelectProductTableProps> = ({
                   {order.currency_code.toUpperCase()}
                 </Table.Cell>
               </Table.Row>
-              {checked && (
+              {checked && !isSwapOrClaim && (
                 <Table.Row className="last:border-b-0 hover:bg-grey-0">
                   <Table.Cell></Table.Cell>
                   <Table.Cell colspan={2}>
```
kasperkristensen, line 220:
```suggestion
                  <Table.Cell colSpan={2}>
```

kasperkristensen (changes requested):
Great job! Added a couple of comments with some slight changes 😄

src/components/organisms/rma-select-receive-product-table/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+                      className={clsx(
+                        "w-5 h-5 flex items-center justify-center rounded cursor-pointer hover:bg-grey-20 ml-2"
+                      )}
+                    >
+                      <PlusIcon size={16} />
+                    </span>
+                  </div>
+                ) : (
+                  <span className="text-grey-40">
+                    {item.quantity - item.returned_quantity}
```
kasperkristensen, line 161: It's possible to make this return a negative number. As an example if you have an line item with a quantity of 4, and you first make a return on two of them and then go to create a new return on a third of them, then the item that is being displayed will look like this: 

```json
...
quantity: 1
refundable: 7300
returned_quantity: 2
shipped_quantity: 4
...
```
Because the quantity here is 1 (the requested return quantity) and there has already been 2 items returned, it will then display `-1`

kasperkristensen (approved):
Awesome LGTM!

PR #540, opened 2022-05-01
feat: price list bulk edit table

fPolic:
**What**
- add the price list bulk edit table

**Depends on**
- https://github.com/medusajs/medusa/pull/1503

**TODO**
- [x] add pagination (design + new ticket?)
- [x] horizontal scroll


zakariaelas (approved):
LGTM! Amazing work Frane 🤩 ! Nailed it 🔨

olivermrbl (approved):
LGTM, really well done @fPolic!

src/domain/pricing/bulk-editor/index.tsx
```diff
@@ -0,0 +1,1072 @@
+import React, { useEffect, useMemo, useRef, useState } from "react"
```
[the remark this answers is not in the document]
fPolic: Thanks @zakariaelas! Happy to hear that you like it 😊

src/domain/pricing/bulk-editor/index.tsx
```diff
@@ -0,0 +1,1072 @@
...
+      const active = isActive(variantId, regionId)
+      toggleActive(cellKey)
+
+      if (active) {
+        const wasFocused = e.target === document.activeElement
+        e.target.blur()
+
+        if (wasFocused) {
+          // find another active cell to focus
+          const id = Object.keys(activeFields).filter((k) => k !== cellKey)[0]
```
pKorsholm, line 613: nit: since we are only looking for an active cell we could use `find` instead to terminate at the first cell that satisfies the condition

src/domain/pricing/bulk-editor/index.tsx
```diff
@@ -0,0 +1,1072 @@
...
+    //         .find((r) => r.id === regionId)!
+    //         .currency_code.toUpperCase()
+    //     ]
+    //
+    //   formatted = (
+    //     Math.round(parseFloat(amount) * 10 ** c.decimal_digits) /
+    //     10 ** c.decimal_digits
+    //   ).toFixed(c.decimal_digits)
+    // }
+
```
pKorsholm, line 748: rm if not to be used in the future

src/domain/pricing/bulk-editor/index.tsx
```diff
@@ -0,0 +1,1072 @@
...
+  /*
+   * A handler that implement logic for arrow keys multi select.
+   */
+  const onKeyDown = (
+    e: React.KeyboardEvent<HTMLInputElement>,
+    variantId: string,
+    regionId: string
+  ) => {
+    if (e.key === "Tab") {
+      e.preventDefault()
```
pKorsholm, line 639: suggestion for future ticket, include tab to go to next cell in row of first cell in next if current cell is last

src/domain/pricing/bulk-editor/index.tsx
```diff
@@ -0,0 +1,1072 @@
...
+/**
+ * Product section container, renders product variant rows.
+ */
+function ProductSection(props: ProductSectionProps) {
+  const {
+    activeRegions,
+    product,
+    isFirst,
+    priceChanges,
+    setPriceChanges,
```
pKorsholm, line 356: remove `priceChanges` and `setPricesChanges` as they are unused

pKorsholm (approved):
LGTM! this is great Frane! 😍 
I have a couple of nits.

One thing (and this might just be my preference) I think it would be cool to create something like a `components` folder in the bulk-editor folder in order to separate the components a bit and to decrease the size of the file.

fPolic:
> One thing (and this might just be my preference) I think it would be cool to create something like a `components` folder in the bulk-editor folder in order to separate the components a bit and to decrease the size of the file.

I find it easier to work on a feature if it is "contained" in a single file (no need to jump between files), but it is a bit harder to review that way 😬. I agree that it would make sense to split the components into separate files in case components are used outside of `bulk-editor`, but since we only export `BulkEditorContainer` I would leave it this way for now.

src/domain/pricing/bulk-editor/index.tsx
```diff
@@ -0,0 +1,1052 @@
...
+    }
+
+    document.addEventListener("mousedown", handler)
+    return () => document.removeEventListener("mousedown", handler)
+  }, [])
+
+  /* ********** HELPERS ********** */
+
+  const activateField = (cellKey: string) => {
+    setActiveFields({ ...activeFields, [cellKey]: true })
```
adrien2p, line 537: should it be `fieldKey` for consistency since everything is a field?

adrien2p: same question for all of them =)

src/domain/pricing/bulk-editor/index.tsx
```diff
@@ -0,0 +1,1052 @@
...
+    } else {
+      activateField(cellKey)
+    }
+  }
+
+  const setActiveCellsFromPointers = () => {
+    const next = {}
+    const [variants, regions] = matrix
+
+    for (let i = pointers.minV; i <= pointers.maxV; i++) {
```
adrien2p, line 559: I was wondering what minV maxV and the same for R was standing for. now I know, would probably be worth naming them differently

src/domain/pricing/bulk-editor/index.tsx
```diff
@@ -0,0 +1,1052 @@
...
+    setActiveFields(next)
+  }
+
+  /* ********** SELECTORS ********** */
+
+  const isActive = (variantId: string, regionId: string) =>
+    activeFields[getPriceKey(variantId, regionId)]
+
+  const getPriceChange = (variantId: string, regionId: string) => {
+    const key = getPriceKey(variantId, regionId)
```
adrien2p, line 576: should try to be consistent, sometimes it is key, sometimes it is cellKey

src/domain/pricing/bulk-editor/index.tsx
```diff
@@ -0,0 +1,1052 @@
...
+    e: React.MouseEvent<HTMLInputElement>,
+    variantId: string,
+    regionId: string
+  ) => {
+    const cellKey = getPriceKey(variantId, regionId)
+
+    if (e.shiftKey) {
+      e.preventDefault() // do not focus
+
+      const active = isActive(variantId, regionId)
```
adrien2p, line 600: active -> isActive

src/domain/pricing/bulk-editor/index.tsx
```diff
@@ -0,0 +1,1052 @@
...
+      toggleActive(cellKey)
+
+      if (active) {
+        const wasFocused = e.target === document.activeElement
+        e.target.blur()
+
+        if (wasFocused) {
+          // find another active cell to focus
+          const id = Object.keys(activeFields).find((k) => k !== cellKey)
+          document.querySelectorAll(`[data-id="${id}"]`)[0].focus()
```
adrien2p, line 610: isn't the id uniq? if it is we should privilege `querySelector` instead?

src/domain/pricing/bulk-editor/index.tsx
```diff
@@ -0,0 +1,1052 @@
...
+      const active = isActive(variantId, regionId)
+      toggleActive(cellKey)
+
+      if (active) {
+        const wasFocused = e.target === document.activeElement
+        e.target.blur()
+
+        if (wasFocused) {
+          // find another active cell to focus
+          const id = Object.keys(activeFields).find((k) => k !== cellKey)
```
adrien2p, line 609: Tell me if I am wrong but here we could focus any active field before or after the one that was currently active? since the activeFields does not hold any order, could that lead to weird behaviour from the end user point of view?

src/domain/pricing/bulk-editor/index.tsx
```diff
@@ -0,0 +1,1052 @@
...
+          document.querySelectorAll(`[data-id="${id}"]`)[0].focus()
+        }
+      } else {
+        e.target.focus()
+      }
+    } else {
+      // TODO: the issue here is that `onAmount` is called by the underlying library on blur (which is triggered after this on click handler)
```
adrien2p, line 616: is it due to the bubbling strategy? so maybe you can prevent the bubbling by using `stopPropagation` and manage those behaviours in the right order (for us) and then use `dispatchEvent` for the click action by giving the ref of the clicked cell?

src/domain/pricing/bulk-editor/index.tsx
```diff
@@ -0,0 +1,1052 @@
...
+
+  /*
+   * A handler that implement logic for arrow keys multi select.
+   */
+  const onKeyDown = (
+    e: React.KeyboardEvent<HTMLInputElement>,
+    variantId: string,
+    regionId: string
+  ) => {
+    if (e.key === "Tab") {
```
adrien2p, line 634: I would suggest to put all the keys in an Enum somewhere to have a quick look at the supported key bindings and to avoid any typo for any usage across the app

src/domain/pricing/bulk-editor/index.tsx
```diff
@@ -0,0 +1,1052 @@
...
+    !!variantsOfList[variantId]
+
+  const checkForChanges = () => {
+    const hasChanges = !!Object.keys(priceChanges).length
+
+    if (hasChanges) {
+      if (
+        confirm(
+          "There are unsaved changes. Do you want to discard and continue?"
+        )
```
adrien2p, line 859: should we move that just before the if and store the returned value?

src/domain/pricing/bulk-editor/index.tsx
```diff
@@ -0,0 +1,1052 @@
...
+    prices.forEach((p) => {
+      delete p.id
+      if (p.region_id) {
+        delete p.currency_code
+      }
+    }) // new records will be created
+
+    updatePriceList.mutate(
+      { prices, override: true },
+      { onSuccess: () => closeEditor() }
```
adrien2p, line 975: Are we sure that we want to close the editor automatically after a save? should we just show a message and let the user decide. I am thinking the case the user just want to save middleway and then continue few minutes later for example (after a break :)). He would have to re open the editor and config the fields to show.

src/domain/pricing/bulk-editor/index.tsx
```diff
@@ -0,0 +1,1052 @@
...
+    )
+  }
+
+  const displayRegions = activeRegions
```
adrien2p, line 979: r, i, a, b represents all an activeRegion but are named differently, goes with the one letter comments.

src/assets/styles/global.css
```diff
@@ -419,3 +419,7 @@
 .remove-number-spinner {
   -moz-appearance: textfield;
 }
+
+.hide-scrollbar::-webkit-scrollbar {
+  display: none;
+}
```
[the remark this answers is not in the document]
fPolic: yep, good catch, we should be inclusive towards the other browsers too :)

src/components/molecules/table-fields-filters/index.tsx
```diff
@@ -0,0 +1,198 @@
...
+  onChange: () => void
+}
+
+/******************** COMPONENTS ********************/
+
+/**
+ * Table field chip component.
+ */
+function Chip(props: ChipProps) {
+  const { remove, short } = props
```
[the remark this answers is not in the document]
fPolic: Could be, but I think it is a bit cleaner this way. This is a case of a React clean code practice (similar to extracting inline code to a variable).

src/components/molecules/table-fields-filters/index.tsx
```diff
@@ -0,0 +1,198 @@
...
+    onChange(selectedFields)
+  }, [selectedFields])
+
+  const removeSelected = (id: string) => {
+    setSelectedFields(selectedFields.filter((f) => f !== id))
+  }
+
+  const visibleFields = selectedFields.map((id) =>
+    fields.find((f) => f.id === id)
+  )
```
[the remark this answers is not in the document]
fPolic: `selectedFields` is a subset of `fileds` array (i.e. a list of ids of the subset), so implicitly, there shouldn't be any nullish values.

src/components/organisms/price-input/index.tsx
```diff
@@ -40,22 +59,40 @@ function PriceInput(props: PriceInputProps) {
       </div>
 
       <AmountField
-        step={step}
```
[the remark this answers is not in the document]
fPolic: It can be `undefined`, in that case, the library skips increment/decrement logic on arrow up/down.

src/domain/pricing/bulk-editor/index.tsx
```diff
@@ -0,0 +1,1052 @@
...
+
+      {activeRegions.map((r) => {
+        const current = variant.prices.find(
+          (p) =>
+            p.region_id === r.id &&
+            p.min_quantity === null &&
+            p.max_quantity === null
+        )
+
+        const a =
```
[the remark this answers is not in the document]
fPolic: I agree in this case, this was intended to be temporary, I was planning to refactor this logic for the amount formating in a helper function.

src/domain/pricing/bulk-editor/index.tsx
```diff
@@ -0,0 +1,1052 @@
...
+          typeof current?.amount === "number"
```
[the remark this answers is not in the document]
fPolic: yes 👍  since this is a MA record, `amount` is always a number

src/domain/pricing/bulk-editor/index.tsx
fPolic: Indeed :)

src/domain/pricing/bulk-editor/index.tsx
fPolic: At this point, a user is editing multiple cells which have a "virtual" focus so it doesn't matter much from the user's point of view which one is actually focused in the DOM.

src/domain/pricing/bulk-editor/index.tsx
fPolic: No, the issue is that the underlying component library is triggering onAmountChange callback when the actual amount didn't change and the callback is setting the current edit value asynchronously.

src/domain/pricing/bulk-editor/index.tsx
fPolic: Hmm a question for Ludvig for sure but since this form behaves as a modal - it is usually expected that modals dismiss if an action is successfully completed.

src/domain/pricing/bulk-editor/index.tsx
adrien2p: Does the callback should call another function first to handle it first and then proceed with the rest then?

src/domain/pricing/bulk-editor/index.tsx
adrien2p: Yeah I agree, but generally, the modal hold a form that you have to fully complete before saving. Here we are in a case where you can do it in multiple time. It is very much a question ^*

src/domain/pricing/bulk-editor/index.tsx
fPolic: Naming is shard :D ... I agree 100% that we should avoid one letter naming in most cases but sometimes it is OK for one liner arrow functions to be an exception to the rule.

adrien2p: We can agree on one level array function? because otherwise we end up with things like a.id === r or r.id and in that case for example I have to go back into the code itself to see what does that means 😂

And even on one level we would have to go back to the code and implementation to see what to means since out of context I have no clue. Typic example

adrien2p: Si lets settle on agree for one liner arrow function 💪

PR #541, opened 2022-05-02
Fix: Rename promotion to discount

pKorsholm:
**What**
- rename promotion to discount
- make valid_for optional in `discountToFormValuesMapper`


olivermrbl (approved):
LGTM!

PR #543, opened 2022-05-03
Fix: Notifications when cancelling exchanges

pKorsholm:
**What**
- notification handling when cancelling returns on exchanges and exchanges 

**How**
- make handling methods async and use the delete prompts for cancelling exchanges and returns

**Why**
- previously when cancelling an exchange without first cancelling the return no error was shown and nothing was cancelled

### DIFF
--- a/src/components/molecules/timeline-events/exchange.tsx
+++ b/src/components/molecules/timeline-events/exchange.tsx
@@ -238,7 +238,7 @@ const Exchange: React.FC<ExchangeProps> = ({ event, refetch }) => {
       {showCancel && (
         <DeletePrompt
           handleClose={() => setShowCancel(!showCancel)}
-          onDelete={async () => handleCancelExchange()}
+          onDelete={handleCancelExchange}
           confirmText="Yes, cancel"
           heading="Cancel exchange"
           text="Are you sure you want to cancel this exchange?"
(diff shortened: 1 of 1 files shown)


src/components/molecules/timeline-events/exchange.tsx
```diff
@@ -112,13 +112,13 @@ const Exchange: React.FC<ExchangeProps> = ({ event, refetch }) => {
...
-    cancelExchange.mutate(event.id)
+  const handleCancelExchange = async () => {
+    await cancelExchange.mutateAsync(event.id)
     refetch()
   }
 
-  const handleCancelReturn = () => {
-    cancelReturn.mutate()
+  const handleCancelReturn = async () => {
+    await cancelReturn.mutateAsync()
```
fPolic, line 121: since this is `async` now, `DeletePrompt.onDelete` prop can accept just  `handleCancelReturn` instead of 
`async () => handleCancelReturn()`

fPolic (approved):

src/components/molecules/timeline-events/exchange.tsx
pKorsholm: Good catch 😄

zakariaelas (approved):
LGTM!

Issue #546, opened 2022-05-03
[bug] UI: Several cases results in error on creating a new variant


kasperkristensen:
**How to reproduce**

For each case do the following:

1. Go to an existing product, and click add variant to open the modal.

Case 1:

As no input fields are marked as required, I try and press save without filling out any fields. I won't be allowed to press save but get no explanation why that is.

This is because my product has an option, such as color, and I haven't filled out the value.

Case 2:

I fill out the option value, but my request fails with a bad request error because title must be string.

Case 3:

I fill out the option value and title, and press save, but once again get a bad request error because inventory_quantity must be a number.

Case 4:

I don't add a price to the variant I am creating resulting in an error because the save method tries to access the array `data.prices` but it does not exist.

PR #548, opened 2022-05-03
fix: bugs related to creating variant

kasperkristensen:
**What**

See #546 for description of errors that are fixed.

**How**

- If no title is provided in the form, create a title from the joining the option values. Same way as we do it when creating a new product w/ variants.
- If not inventory quantity is provided add `inventory_quantity: 0` to the request.
- Add a required `*` to option inputs to signify that they are required to fill out.
- If no prices are entered, focus the `Add price` button, also add required `*` to the Prices label to signify that it is required to add a price.

### DIFF
--- a/src/domain/products/details/variants/variant-editor.tsx
+++ b/src/domain/products/details/variants/variant-editor.tsx
@@ -12,6 +12,7 @@ import Select from "../../../../components/molecules/select"
 import CurrencyInput from "../../../../components/organisms/currency-input"
 import { convertEmptyStringToNull } from "../../../../utils/convert-empty-string-to-null"
 import { countries as countryData } from "../../../../utils/countries"
+import { focusByName } from "../../../../utils/focus-by-name"
 import usePricesFieldArray from "../../product-form/form/usePricesFieldArray"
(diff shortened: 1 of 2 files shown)


src/domain/products/details/variants/variant-editor.tsx
```diff
@@ -79,6 +79,19 @@ const VariantEditor = ({
   }, [variant, store])
 
   const handleSave = (data) => {
+    if (!data.prices) {
+      const element = document.getElementById("add-price")
+      if (element) {
+        element.focus()
+      }
```
fPolic, line 86:
```suggestion
      document.getElementById("add-price")?.focus()
```

fPolic (approved):

src/domain/products/details/variants/variant-editor.tsx
pKorsholm: We also have a utility for focussing by name we could use. Up to you.
```suggestion
      focusByName("add-price")
```

pKorsholm (approved):
lgtm

Issue #554, opened 2022-05-08
Clicking on 'X' icon doesn't close search modal


aliseidev:
![image](https://user-images.githubusercontent.com/64712107/167300347-a8fceb8c-2175-455a-ae9d-372538fc741d.png)

I've checked the code and the icon doesn't have any click handler.

`C:\Users\alisb\Desktop\projects\medusa-admin-dev\src\components\templates\search-modal\index.tsx`

There is already the closeHandle, i'll fix it quickly

aliseidev:
Okay it wan't a bug sorry for the useless issue! 

See: https://github.com/medusajs/admin/pull/555#issuecomment-1121555906

PR #555, opened 2022-05-08
Fix: 'X' icon in searchmodal not working

aliseidev:
Fixes the issue #554

'X' icon not working due to no onClick handle provided. Fixed by providing already existing 'handleClose' prop

### DIFF
--- a/src/components/templates/search-modal/index.tsx
+++ b/src/components/templates/search-modal/index.tsx
@@ -102,7 +102,7 @@ const SearchModal = ({ handleClose }) => {
                   onClick={handleClear}
                   content="Clear search"
                 >
-                  <CrossIcon className="flex text-grey-50" />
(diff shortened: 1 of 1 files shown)


zakariaelas:
Hello @aliseidev, thanks a lot for the contribution 💜 ! 

Unfortunately, the 'X' icon is used for clearing the search query, as described by this tooltip [here](https://github.com/medusajs/admin/pull/555/files#diff-35fbc76a26613e186534649b2495fa183a04fcf3d38a9737fb2291bca2a8231eR100-R104). However, I do think that you're onto something here .. I must admit that the 'X' icon very much looks like a close icon 😅 . We'll try to run this by our designer to get his thoughts on this. 

Closing for now, but will create an issue until we come up with a solution. I'll also tag you on the issue in case you want to take the lead on this one 👍 . Thanks again!

Issue #556, opened 2022-05-09
bug(orders): order date displays incorrectly in detail view


srindom:
It is likely the formatting string here: https://github.com/medusajs/admin/blob/master/src/domain/orders/details/index.tsx#L314 that is incorrect.

PR #557, opened 2022-05-09
Fix: Date format in order details

twosaturdayscode:
Issue: #556 
As issue states it was just about wrong formatting string, for reference see: https://momentjs.com/docs/#/parsing/string-format/

Now it prints the date as expected.


src/components/templates/search-modal/index.tsx
```diff
@@ -102,7 +102,7 @@ const SearchModal = ({ handleClose }) => {
                   onClick={handleClear}
                   content="Clear search"
                 >
-                  <CrossIcon className="flex text-grey-50" />
```
kasperkristensen, line 105: Hi @aliseidev, 

Thanks for the PR, and sorry for taking so long to review it. If you could please remove this onClick you've added. as the button is not supposed to close the search modal, but just clear the text field. Otherwise, it looks great 😄

kasperkristensen (changes requested):

kasperkristensen:
Solved in #637

PR #559, opened 2022-05-11
fix: RMA menus (return request/receive, swap, claim)

kasperkristensen:
**What**

- Fixes issue where items would sometimes not show up in product tables in receive menus.
- Adds types to the components, and also some related components.
- Fixes typing of `IndeterminateCheckbox` component
- Adds typing to Claim and SwapMenu

**How**
- Updates memorized function `allItems` to retrieve all products that are available when creating a return request.
  - If a user created a return request, then canceled it, and made a new request with the same items, they would not appear in the receive menu. This was due to how products were filtered, which prevented an item that was associated with a canceled return to be shown, even if it was now part of an active return.
  - Items that were added to an order through `additional_items` on both `claim_orders` and `swaps` were not shown either.
- Updates filtering function in `RMASelectProductTable`


srindom (commented):
LGTM! 🎉

src/domain/orders/details/swap/create.tsx
```diff
@@ -58,17 +57,14 @@ const SwapMenu = ({ order, onCreate, onDismiss, notification }) => {
...
-      .then(({ data }) => {
-        setShippingOptions(data.shipping_options)
-        setShippingLoading(false)
-      })
-  }, [])
+  const {
+    shipping_options: shippingOptions,
+    isLoading: shippingLoading,
+  } = useAdminShippingOptions({
+    // @ts-ignore TODO remove once #1493 is merged
```
olivermrbl, line 64: comment: @kasperkristensen 1493 has been merged

olivermrbl (approved):
LGTM! 

Single comment to address otherwise good to go! 🥳

kasperkristensen:
@olivermrbl fyi I noticed some strange behaviour in the SwapMenu, that I fixed, and while I was at it I also added types to the Claim and SwapMenu, so I've added a bit more to the PR

package.json
```diff
@@ -11,8 +11,8 @@
   "dependencies": {
     "@emotion/core": "^10.0.28",
     "@emotion/styled": "^10.0.27",
-    "@medusajs/medusa": "^1.3.0",
-    "@medusajs/medusa-js": "^1.2.0",
+    "@medusajs/medusa": "^1.3.1-next.27",
+    "@medusajs/medusa-js": "^1.2.1-next.27",
```
olivermrbl, line 15: issue: Should be left unchanged for now

package.json
```diff
@@ -49,7 +49,7 @@
     "iso8601-duration": "^1.3.0",
     "jsonwebtoken": "^8.5.1",
     "lodash": "^4.17.21",
-    "medusa-react": "^0.3.0",
+    "medusa-react": "^0.3.1-next.27",
```
olivermrbl, line 52: issue: Should be left unchanged for now

olivermrbl (approved):
LGTM! Couple of required changes before merging 👍

PR #568, opened 2022-05-19
fix: price list overrides does not fire request

zakariaelas:
### What 

- price list overrides modal did not fire a request when the "save and close" button is clicked

### How

- add mutation from `medusa-react`
- fixed behavior of checkboxes when selecting variants


olivermrbl (approved):

olivermrbl:
@zakariaelas Is this ready to be merged? 💯

zakariaelas:
@olivermrbl would like to push one extra commit for filtering the prices in the modal (will add some more context later ✌️ )

PR #572, opened 2022-05-24
fix: prices length in prices overrides modal

zakariaelas:
### What

- remove console.log
- filtered prices length should be reflected in the product variant tree

### DIFF
--- a/src/domain/pricing/details/sections/prices-details/edit-prices-overrides/index.tsx
+++ b/src/domain/pricing/details/sections/prices-details/edit-prices-overrides/index.tsx
@@ -80,6 +80,7 @@ const EditPricesOverridesModal = ({
                     key={variant.id}
                     onClick={getOnClick(variant)}
                     variant={variant}
+                    prices={variant.prices.filter((pr) => pr.price_list_id)}
                   />
                 </CollapsibleTree.Leaf>
               ))}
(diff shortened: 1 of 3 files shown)


src/domain/pricing/details/sections/prices-details/edit-prices-overrides/index.tsx
```diff
@@ -80,6 +80,7 @@ const EditPricesOverridesModal = ({
                     key={variant.id}
                     onClick={getOnClick(variant)}
                     variant={variant}
+                    prices={variant.prices.filter((pr) => pr.price_list_id)}
```
olivermrbl, line 83: question: In case the variant is part of more than one price list, would we not end up showing prices for all of those in the modal with the current filtering?
```suggestion
                    prices={variant.prices.filter((pr) => pr.price_list_id === priceListId )}
```

olivermrbl (approved):
LGTM, a single question, that I'd like you to address before merging 👍

src/domain/pricing/details/sections/prices-details/edit-prices-overrides/index.tsx
zakariaelas: correct me if I'm wrong @olivermrbl but for me it seems like that never happens. Even when a variant is part of more than one price list, what we end up with is default prices + prices defined in that particular price list. Does it behave in a different way for you?

olivermrbl: Got it! My bad :)

PR #573, opened 2022-05-24
fix: create price list form

zakariaelas:
### What

- remove hard coded variant value from form mapper
- use price overrides modal to edit prices when creating a price list

### DIFF
--- a/src/domain/pricing/details/sections/prices-details/edit-prices-overrides/index.tsx
+++ b/src/domain/pricing/details/sections/prices-details/edit-prices-overrides/index.tsx
@@ -11,6 +11,7 @@ import PriceOverrides from "../../../../../../components/templates/price-overrid
 import ProductVariantLeaf from "./product-variant-leaf"
 import { useParams } from "@reach/router"
 import { mapToPriceList } from "./mappers"
+import { mergeExistingWithDefault } from "../../../utils"
 
 type EditPricesOverridesModalProps = {
   product: Product
@@ -121,15 +122,6 @@ const EditPricesOverridesModal = ({
       </Modal.Body>
     </LayeredModal>
   )
-}
-
-const mergeExistingWithDefault = (variantPrices: any[] = [], defaultPrices) => {
-  return defaultPrices.map((pr) => {
-    const price = variantPrices.find(
-      (vpr) => vpr?.currency_code === pr.currency_code
-    )
-    return price || pr
-  })
 }
 
 export default EditPricesOverridesModal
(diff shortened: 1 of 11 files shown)


olivermrbl (changes requested):
_Note: The following might not be directly related to this PR, so please let me know if you think we should create separate issues_

Currently experiencing strange behavior when submitting an update. Will attach a screen recording below.

1. When editing a variant without price list prices, I am not allowed to add prices
2. When setting "Apply to all..." it only really applies to a single variant out of all

Both scenarios can be found in the recording.


https://user-images.githubusercontent.com/59018053/170210588-2d7874ac-a24b-49c3-860e-07661f354ec2.mov

zakariaelas:
To address the things that look weird on the recording:
1. The product variant tree shows an incorrect number of prices. For example, in the video, it showed 4 prices, but when clicked, no prices appeared. This is fixed in #572.
2. When editing a variant without price list prices, I am not allowed to add prices: This was the normal behavior of the price overrides modal, but I will add the same workaround I did for the create price list form i.e: show a price input for every store currency
3. The price list overrides modal is quite tricky in an editing scenario and I think we should settle on how we want it to behave. To explain how it can be tricky:
    - Variant A has 3 prices set (usd, eur, dkk)
    - Variant B has 5 prices set (usd, eur, dkk, sek, gbp)
    - When editing Variant B, I also check the checkbox for Variant A to signal that I also want these prices set for Variant A
    - Now the update for Variant A is no longer just a simple update, because now, we have to override the usd, eur, dkk prices, and create new ones for sek and gpb. If we just send the prices to the `POST /admin/price_lists/:id` route just like we get them from the form, variant A will end up with 8 prices (3 existing ones + 5 new ones). We could try to consolidate all this in our frontend code using the `currency_code` for example, but imo this is not super ideal and becomes even more confusing when we introduce the notion of regional prices. 

Wdyt?

src/domain/pricing/details/sections/prices-details/edit-prices-overrides/index.tsx
```diff
@@ -112,4 +123,13 @@ const EditPricesOverridesModal = ({
...
 }
 
+const mergeExistingWithDefault = (variantPrices: any[] = [], defaultPrices) => {
+  return defaultPrices.map((pr) => {
+    const price = variantPrices.find(
+      (vpr) => vpr?.currency_code === pr.currency_code
+    )
+    return price || pr
+  })
+}
```
olivermrbl, line 133: comment: This method is defined in two files. Should we move to a util?

olivermrbl (approved):
Great work @zakariaelas ! A comment but otherwise LGTM

src/domain/pricing/details/sections/prices-details/edit-prices-overrides/index.tsx
zakariaelas: nice catch! yes totally!

PR #577, opened 2022-05-25
chore: Clean up unused dependencies

olivermrbl:
**What**
Upgrade Storybook dependency + configuration.

Removes dependencies:
- `@emotion/styled`
- `@emotion/core`
- `@rebass/forms`
- `gatsby-plugin-portal`
- `react-collapse`
- `rebass`
- `prop-types`
- `@medusa/medusaj-js`
- `@popperjs/core`
- `jsonwebtoken`
- `react-dropzone`
- `react-tooltip`
- `react-multi-select-component`

Removes all theme files related to Emotion

_Note: Used npm utility `depcheck` to find unused dependencies_


zakariaelas (approved):
LGTM! 

Building storybook seems to fail, not entirely sure where the issue is coming from. Did we upgrade any of our dependencies in a previous PR?

package.json
```diff
@@ -5,15 +5,11 @@
   "version": "0.1.0",
   "author": "@medusajs",
   "resolutions": {
-    "devcert": "1.2.0",
+    "devcert": "1.2.1",
```
olivermrbl, line 8: comment: 1.2.0 had a security vulnerability

src/components/molecules/select/index.tsx
```diff
@@ -318,72 +316,63 @@ const SSelect = React.forwardRef(
...
-          >
-            {
-              <GetSelect
-                isCreatable={isCreatable}
-                searchBackend={filterOptions}
-                options={
-                  hasSelectAll && isMultiSelect
-                    ? [{ value: "all", label: "Select All" }, ...options]
-                    : options
+          {
```
olivermrbl, line 319: comment: Looks more dangerous than it is. Simply removing the `CacheProvider` from Emotion

olivermrbl:
> Building storybook seems to fail, not entirely sure where the issue is coming from. Did we upgrade any of our dependencies in a previous PR?

Seems to be an issue with a version mismatch between Gatsby and Storybook. As per Gatsby's recommendation, I've updated Storybook and its configuration to the latest version.

PR #529, opened 2022-04-21
feat: price list details prices overrides

zakariaelas:
### What

- adds prices overrides screen
- adds edit price overrides modal which uses the prices overrides component
- adds a `MedusaPriceInput` component which augments the `PriceInput` component with extra functionality
- adds a `MedusaPriceInput` story
- adds a `useToggleState` (for a lack of better name) hook to be used whenever we have a boolean state used to display/hide/toggle something (i.e: whenever we use state to show a model, we create at least 2 functions for opening/closing the modal)

PR #440, opened 2022-03-17
feat: `ProductVariantTree` component

kasperkristensen:
**What**
- Implements `ProductVariantTree` component that should be used in new PriceList UI.
- Actions are todo and should be updated once implemented in the actual context.

https://user-images.githubusercontent.com/45367945/158795356-ef77f861-0577-43f8-b6b0-3d004514b3a3.mov

closes #436

PR #578, opened 2022-05-26
feat: price list details

zakariaelas:
### What

- adds price list details domain functionality to medusa admin

### How

- #573
- #568
- #529
- #523
- #511
- #440


src/components/atoms/file-upload-field/index.tsx
```diff
@@ -5,16 +5,25 @@ type FileUploadFieldProps = {
...
   filetypes: string[]
   errorMessage?: string
-  placeholder?: string
+  placeholder?: React.ReactElement | string
   className?: string
+  text?: React.ReactElement | string
 }
 
+const defaultText = (
+  <span>
```
adrien2p, line 14: question: does {’ ‘} is equivalent to &nbsp; ? Because i think this one can be break and the other one is unbreakable and I tend to use the second one

olivermrbl (commented):
When trying to "Edit manually", I am getting the following error:
image

image

Seems to be an issue with `usePriceListForm` in component `<ProductPrices />`

zakariaelas:
@olivermrbl fixed in #579 .. let's make sure to get that one in before merging this one 👍

package.json
```diff
@@ -11,7 +11,7 @@
   "dependencies": {
     "@emotion/core": "^10.0.28",
     "@emotion/styled": "^10.0.27",
-    "@medusajs/medusa": "^1.2.1",
+    "@medusajs/medusa": "^1.1.65-alpha.63",
```
olivermrbl, line 14: issue: Should be left unchanged

zakariaelas: @olivermrbl fixed 👍

PR #579, opened 2022-05-27
fix: price list details edit manually

zakariaelas:
### What

- fixes the "edit manually" feature

### How

- The `<ProductPrices />` component was using the form hook without being wrapped inside a `PriceListFormProvider`. 
- Moved logic around so that the `EditPrices` component also works as expected (while waiting for the bulk editor to come in)

### DIFF
--- a/src/components/organisms/product-variant-tree/index.tsx
+++ b/src/components/organisms/product-variant-tree/index.tsx
@@ -11,6 +11,7 @@ type LeafProps = {
     id: string
     currency_code: string
     amount: number
+    price_list_id: string | null
   }[]
 }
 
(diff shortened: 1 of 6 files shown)


olivermrbl (commented):
I am seeing two different num prices for the variants.

**Edit price overrides modal**
image

**Edit manually view**
image

src/components/organisms/product-variant-tree/index.tsx
```diff
@@ -62,9 +64,9 @@ const ProductVariantLeaf = ({ sku, title, prices = [] }: LeafProps) => {
       </div>
       <div className="flex items-center text-grey-50 flex-1 justify-end">
         <div className="text-grey-50 mr-xsmall">
-          {prices.length ? (
+          {filteredPrices.length ? (
             <span>{`${prices.length} price${
```
olivermrbl, line 68: comment: Found the culprit 😄 
```suggestion
            <span>{`${filteredPrices.length} price${
```

olivermrbl (changes requested):

src/components/organisms/product-variant-tree/index.tsx
zakariaelas: haha shit nice catch 😆

olivermrbl (approved):

PR #618, opened 2022-06-27
feat: Order export

pKorsholm:
**What**
- order export button

note: this depends on core pr: https://github.com/medusajs/medusa/pull/1603


olivermrbl (approved):
LGTM!

src/domain/orders/index.tsx
```diff
@@ -1,41 +1,95 @@
...
 const OrderIndex: React.FC<RouteComponentProps> = () => {
   const view = "orders"
 
+  const createBatchJob = useAdminCreateBatchJob()
+  const notification = useNotification()
+
+  const {
+    open: openExportModal,
+    close: closeExportModal,
+    state: exportModalOpen,
```
adrien2p, line 27: suggsestion: `isExportModalOpen`

src/domain/orders/index.tsx
```diff
@@ -1,41 +1,95 @@
...
+    createBatchJob.mutate(reqObj, {
+      onSuccess: () => {
+        notification("Success", "Successfully initiated export", "success")
+      },
+      onError: (err) => {
+        notification("Error", getErrorMessage(err), "error")
+      },
+    })
+
+    closeExportModal()
```
adrien2p, line 59: question: Since we are not awaiting for the creation, could it be possible that the modal is closed before the notification is displayed? is that intended? Maybe that we should call the closeExportModal on success?

adrien2p (approved):
LGTM! only 2 questions/suggestions ;)

olivermrbl:
This should be good to go @adrien2p ?

PR #628, opened 2022-07-08
Feat(admin): feature flags

pKorsholm:
**What**
- Feature flags for admin ui components

**Why**
- Featureflags corresponding to flags set in `core` will allow us to mask ui elements and behavior that are not ready to be used by end-users and are still part of features being developed in `core`. 

**How**
- Add a feature-flag provider to the component tree in gatsby browser
  - ff-provider fetches the store and active feature flags in order to get active feature flags for the feature toggle provider from this library: https://www.npmjs.com/package/react-feature-toggles

dependent on: https://github.com/medusajs/medusa/pull/1819
### How to use the feature flags in admin

In order to use feature flags in admin, the component `FeatureToggle` is used as seen below. The feature flag used is the same as in `core`.
 
```jsx
import FeatureToggle from "../../components/fundamentals/feature-toggle"

...

const Test = () => {
  return (
    <FeatureToggle featureFlag="sales_channels">
      <Router>
        <Index path="/" />
        <Details path=":id" />
      </Router>
    </FeatureToggle>
  )
}
```

or inline in components: 

```tsx
const { isFeatureEnabled } = React.useContext(FeatureFlagContext)

if(isFeatureEnabled("sales_channels")){
  ...
}
```

### DIFF
--- a/src/components/fundamentals/feature-toggle.tsx
+++ b/src/components/fundamentals/feature-toggle.tsx
@@ -11,9 +11,9 @@ const FeatureToggle: React.FC<FeatureToggleProps> = ({
   showOnlyWhenDisabled = false,
   children,
 }) => {
-  const { featureToggleList } = React.useContext(FeatureFlagContext)
+  const { isFeatureEnabled } = React.useContext(FeatureFlagContext)
 
-  const showContent = !!featureToggleList[featureFlag] === !showOnlyWhenDisabled
+  const showContent = isFeatureEnabled(featureFlag) === !showOnlyWhenDisabled
   return showContent ? <>{children}</> : null
 }
 
--- a/src/components/fundamentals/feature-toggle.tsx
+++ b/src/components/fundamentals/feature-toggle.tsx
@@ -1,0 +1,20 @@
+import React from "react"
+import { FeatureFlagContext } from "../../context/feature-flag"
+
+export type FeatureToggleProps = {
+  featureFlag: string
+  showOnlyWhenDisabled?: boolean
+}
+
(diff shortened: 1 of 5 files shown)


src/context/feature-flag.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  useEffect(() => {
+    if (isFetching || !store || !isLoggedIn) {
+      return
+    }
+
+    setFeatureFlags(store["feature_flags"])
+  }, [isFetching, store, isLoggedIn])
+
+  const featureToggleList = featureFlags.reduce(
```
olivermrbl, line 32: issue(blocking):

The implementation of this context does not take into account our revised backend implementation of the feature flags response which is:
```ts
[ { "key": "sales_channels", "value": true }, { "key": "multi-warehouse", "value": false} ]
```

olivermrbl (changes requested):

olivermrbl (approved):
LGTM - super nice addition!

@fPolic @pKorsholm Please incorporate feature flags in #647 and #635 💪

src/components/fundamentals/feature-toggle.tsx
```diff
@@ -0,0 +1,20 @@
+import React from "react"
+import { FeatureFlagContext } from "../../context/feature-flag"
+
+export type FeatureToggleProps = {
+  featureFlag: string
+  showOnlyWhenDisabled?: boolean
```
adrien2p, line 6: question: Does that allow to display a component only if the flag is disabled?

adrien2p (approved):
LGTM 🎉

fPolic (approved):
LGTM! 🎉

PR #631, opened 2022-07-09
Feat/product-collection-images

edihasaj:
After adding this: https://github.com/medusajs/medusa/pull/1799, I've worked on implementing the front end for that based on the philosophy of other components created.

Basically this functionality has been added:
a) migrate from modal view to fixed page view (for create and update)
b) ability to upload new images.


edihasaj:
old_col
new_col
old_detail_col
new_detail_col

Issue #625, opened 2022-07-02
Adding initial options and variants for product doesn't work


chemicalkosek:
Steps to reproduce

1. Start creating a new product
2. Start adding options and variants from the ui attached in the screenshot below
![image](https://user-images.githubusercontent.com/11018707/176994639-45f34ee9-2304-4bc5-9473-3c4475d45ddb.png)
3. Try to save the product
4. See all the variants are gone. Only default variant exists.

olivermrbl:
Thanks for reporting this one in @chemicalkosek. We are currently revisiting our forms across the entire admin system and you can expect a fix for this issue to be part of an upcoming PR 👍

chemicalkosek:
Btw I think the Tags label on this input is a reminiscent of reused Input from Product Tags. 
It should probably be named Option Values or something like that (?).

olivermrbl:
Solved in #636

PR #636, opened 2022-07-14
fix: product creation with variants

fPolic:
**What**
- fix an issue where products couldn't be created with initial variants

**How**
- by fixing wrong memoization - the function that provided variant data upon submission had stale closure

fixes https://github.com/medusajs/admin/issues/625


olivermrbl (approved):
Amazing! Great catch 🙌

PR #644, opened 2022-07-17
Fix customer detail page

revskill10:
- Fix some `undefined` error in customer page


ymaheshwari1:
I think there is no need to push yarn.lock and package-lock.json files.

olivermrbl:
Fixed in #649

PR #676, opened 2022-08-24
feat: variant section revamp

kasperkristensen:
Fixes CORE-423, CORE-399, CORE-384


src/domain/products/components/customs-form/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+          pattern: FormValidator.whiteSpaceRule("HS Code"),
+        })}
+        errors={errors}
+      />
+      <Controller
+        name={path("origin_country")}
+        control={control}
+        render={({ field }) => {
+          return (
+            <Select
```
olivermrbl, line 61: todo: Select component is off in both its states:

**Closed**
image

**Open**
image

src/domain/products/components/variant-form/index.tsx
```diff
@@ -0,0 +1,220 @@
...
+ * }
+ */
+const VariantForm = ({ form }: Props) => {
+  const { fields } = useFieldArray({
+    control: form.control,
+    name: "options",
+  })
+
+  return (
+    <Accordion type="multiple" defaultValue={["general"]}>
```
olivermrbl, line 63: question: Upon expanding an accordion, do we want all others to be closed? This is not the current behaviour but I believe that is the conventional definition of an accordion - please correct me if I am wrong.

We could have an option for the accordion component defining whether or not multiple can be open at the same time.

src/domain/products/edit/sections/variants/edit-variant-modal.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+  const {
+    onUpdateVariant,
+    onAddVariant,
+    addingVariant,
+    updatingVariant,
+  } = useEditProductActions(product.id)
+
+  const onSubmit = handleSubmit((data) => {
+    if (isDuplicate) {
```
olivermrbl, line 48: todo: Upon duplicating a product, you cannot save and close before opening the pricing accordion at least once. Ideally, the duplicated product variant will inherit prices as well, so you can do "Duplicate" -> "Save". Wdyt?

olivermrbl (commented):
Great job @kasperkristensen! Overall functionality is working seamlessly - only have a couple of comments 💪

src/domain/products/components/variant-form/index.tsx
kasperkristensen: We have that prop on the component, I set it to allow multiple open right now, but we can definetly change it. I personally think it would be a bit annoying, but fine with it if you think its the way to go :)

olivermrbl: Ahh got it! Sorry, missed that. Think it's fine this way 👍

src/domain/products/components/customs-form/index.tsx
kasperkristensen: So this happends with the current select component when you open it by clicking on the arrow button, and it is also the behaviour in the live version of admin. If you click on the actual input then it should open up the list. Will try to resolve some of these issues + the design of the select in a seperate PR.

src/domain/products/edit/sections/variants/edit-variant-modal.tsx
kasperkristensen: Fixed, had the save button disabled while the form was not dirty, this is of course not the correct behavior for a duplicate situation. Should be good now!

Issue #574, opened 2022-05-25
Move DeclarativeToaster to atoms folder


olivermrbl:
**What**
Component which are used but not grouped according to the components structure of the folder

**How**
- Move components/declarative-toaster to atoms
- Update imports

PR #681, opened 2022-09-06
chore: Move DeclarativeToaster to atoms folder

djgovani:
What

- Component which are used but not grouped according to the components structure of the folder

How

- Move components/declarative-toaster to atoms
- Update imports

Closes https://github.com/medusajs/admin/issues/574


kasperkristensen (approved):
LGTM thanks for your contribution!

PR #682, opened 2022-09-06
Feat/update rhf version

kasperkristensen:


pKorsholm (commented):
`Components/atoms` lgtm

src/components/atoms/back-button/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+import React from "react"
+import ArrowLeftIcon from "../../fundamentals/icons/arrow-left-icon"
+
+type Props = {
+  path?: string
+  label?: string
+  className?: string
+}
+
+const BackButton = ({ path, label = "Go back", className }: Props) => {
```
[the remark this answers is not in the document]
kasperkristensen: Yep, there are multiple instances of this, plus already existing different version of components with the same functionality such as the Currency/Amount inputs. I've created a issue for this, as we need to realign the design across admin, and do some house cleaning of our components to ensure maintainability moving forward. CORE-527

pKorsholm (commented):
customers: lgtm

src/utils/extract-customer-name.ts
```diff
@@ -0,0 +1,69 @@
+import { Cart, Order } from "@medusajs/medusa";
```
adrien2p, line 1: nit(non blocking): we could simplify that with something along that line: wdyt? It is really a proposal and not something to necessary do, it also allow to update the props on which to look for pretty easily and to centralise the logic which is duplicate for each prop. just wanted to get your opinion :)
```ts
const extractCustomerName = (obj?: Cart | Order): string => {
  if (!obj) {
    return "N/A"
  }

  let firstName = ""
  let lastName = ""

  ;["customer", "shipping_address", "billing_address"].some((prop) => {
    firstName = obj[prop]?.first_name
    lastName =  obj[prop]?.last_name
    return !!(lastName || firstName)
  })

  if (firstName && lastName) {
    return `${firstName} ${lastName}`
  }

  if (firstName || lastName) {
    return firstName ?? lastName
  }

  if (obj.email) {
    return obj.email
  }

  return "N/A"
}
}

export default extractCustomerName
```

src/utils/form-validator.ts
```diff
@@ -0,0 +1,19 @@
+import { ValidationRule } from "react-hook-form"
+
+const FormValidator = {
+  whiteSpaceRule: (name: string) =>
+    ({
+      value: /^[^\s]+(?:$|.*[^\s]+$)/,
+      message: `${name} cannot have leading or trailing spaces, or be an empty string.`,
+    } as ValidationRule<RegExp>),
+    nonNegativeNumberRule: (name: string) => ({
```
adrien2p, line 9: nit(non blocking): format error, there is some spaces on the left.

src/utils/nested-form.ts
```diff
@@ -0,0 +1,83 @@
...
+ *   
+ *   return (
+ *     <div>
+ *       <Input {...register(path("city"))} label="City" /> // path("city") resolves as "shipping_address.city" or "billing_address.city" depending on the second argument passed to nestedForm
+ *       <Input {...register(path("postal_code"))} label="Postal Code" />
+ *     </div>
+ *   )
+ * }
+ */
+export function nestedForm<TValues extends FieldValues>(
```
adrien2p, line 51: suggestion: could we simpliy with the following?
```ts
export function nestedForm<TValues extends FieldValues = any, TPath extends FieldPath<TValues> = any>(
  form: UseFormReturn<TValues> | NestedForm<TValues>,
  path?: TPath | string | number
): NestedForm<TValues> {
  // ... 
}
```

src/utils/payment-providers-mapper.ts
```diff
@@ -1,10 +1,27 @@
-export default function (provider) {
+import { Option } from "../types/shared"
+
+export default function (provider: string): Option {
```
adrien2p, line 3: suggestion: I suggest the following simplification which could also allow to extract the providers elsewhere if needed and also simplify to add/remove or update the providers. wdyt?
```ts
const providerDescriptors: { [providerKey: string]: Option } = {
  stripe: {
    label: "Stripe",
    value: "stripe",
  },
  "stripe-ideal": {
    label: "Stripe Ideal",
    value: "stripe-ideal",
  },
  //...
}

export default function (provider: string): Option {
  return providerDescriptors[provider] ?? {
    label: provider,
    value: provider,
  }
}
```

adrien2p (commented):
Nice work 🎉  `Utils` review

src/utils/payment-providers-mapper.ts
adrien2p: it would make it easily configurable and avoid having to update the core function each time we want to update the providers

src/pages/reset-password.tsx
```diff
@@ -114,18 +114,14 @@ const ResetPasswordPage = ({ location }) => {
                 <SigninInput
```
adrien2p, line 114: suggestion: Got some type issues with the token email, wdyt about the following instead?
```ts
let token: { email: string } | null = null
// ... ^ Previously Object type which means that email does not exists on Object
const email = token?.email || parsed?.email as string || ""
```

adrien2p (commented):
Super nice, pages review

src/utils/nested-form.ts
kasperkristensen: Nope, unfortunatly the different overrides are nescessary for this to comply with RHF types. So we have to keep this beauty of a mess 😅

src/hooks/use-debounce.ts
```diff
@@ -1,6 +1,6 @@
 import { useEffect, useState } from "react"
 
-export function useDebounce(value, delay) {
+export const useDebounce = <T extends unknown>(value: T, delay: number): T => {
```
adrien2p, line 3: suggestion: Would it make sense to do the following instead? since they are pretty much equivalent (except some cases) I think it simplify the hook
```ts
export const useDebounce = <T>(value: T, delay: number): T => {
```

adrien2p:
Screenshot 2022-09-07 at 15 49 10

For some reaons, depending on the amount I put in the price input, I get an error 500

adrien2p:
Screenshot 2022-09-07 at 15 54 42

I didn't understand what was that value until I change the visibility, should we indicate that it correspond to the visibility? we could add a popoever that say `visible in the store` or `Not visible in the store`?

adrien2p:
Screenshot 2022-09-07 at 15 59 16

When we try to click on the cross of the selected countries, I was expected that it would deselect all of them, but it just open the select input. is that the actual behaviour?

kasperkristensen:
> Screenshot 2022-09-07 at 15 57 07
> 
> When adding countries to a region, I would have expect to see the countries I have selected. otherwise we have to go through the list too see which ones are selected :) wdyt? Maybe something similar to the sales channel design?

Agreed, had the talk with Ludvig as well, and it's something he is going to revisit as this is not the nicest UX 😄

adrien2p:
Screenshot 2022-09-07 at 16 00 35

Should we display the region such as `name (id)` so that it is easier to find out which one already contains the country?

kasperkristensen:
> Screenshot 2022-09-07 at 15 49 10
> 
> For some reasons, depending on the amount I put in the price input, I get an error 500. With more reasonable amount it works.
> 
> Also, we should maybe disallow a price that is greater than the max sub total no?

Min/Max. subtotal refers to the subtotal of a cart. So these requirements specify that the shipping option is available to carts where the subtotal is >= Min. subtotal and <= Max. subtotal, so it's not related to the price of the shipping option. But interesting with the unknown error on large amounts, but I guess that should be opened as a bug against the core.

src/utils/nested-form.ts
adrien2p: alright, thanks for the clarification 👍

adrien2p:
> > Screenshot 2022-09-07 at 15 49 10
> > For some reasons, depending on the amount I put in the price input, I get an error 500. With more reasonable amount it works.
> > Also, we should maybe disallow a price that is greater than the max sub total no?
> 
> Min/Max. subtotal refers to the subtotal of a cart. So these requirements specify that the shipping option is available to carts where the subtotal is >= Min. subtotal and <= Max. subtotal, so it's not related to the price of the shipping option. But interesting with the unknown error on large amounts, but I guess that should be opened as a bug against the core.

Understood, sorry for the miss-understanding  here.
You are right for the bug in the core, @olivermrbl do you have the logs otherwise we can try it locally

src/types/shared.ts
```diff
@@ -26,3 +26,21 @@ export type PaginationProps = {
 }
 
 export type Idable = { id: string; [x: string]: any }
+
+export type Role = {
+  value: "admin" | "member" | "developer"
```
fPolic, line 31: suggestion: extract to enum

fPolic (commented):
lgtm: hooks ✅ 
lgtm: pages ✅ 
lgtm: types ✅ 
lgtm: utils ✅

src/domain/settings/regions/components/region-form/use-store-data.tsx
```diff
@@ -0,0 +1,58 @@
...
+    )
+  }, [fulfillment_providers, isLoading])
+
+  const currencyOptions = useMemo(() => {
+    if (isLoading) {
+      return []
+    }
+
+    return (
+      store?.currencies.map((c) => ({
```
adrien2p, line 40: question: Should we move that as part of the hook? The idea would be to do it only once when the store is loaded and not have to do it everywhere we need that. the hook could return storeFormattedCurrencies?

src/domain/settings/regions/components/region-form/use-store-data.tsx
```diff
@@ -0,0 +1,58 @@
...
+        label: c.name,
+        value: c.code,
+      })) || []
+    )
+  }, [store, isLoading])
+
+  const countryOptions = countries.map((c) => ({
```
adrien2p, line 47: comment: Same as for the currencies

src/domain/settings/regions/edit/general-section/index.tsx
```diff
@@ -0,0 +1,207 @@
...
+                    )}
+                  </div>
+                ) : (
+                  <p>No countries configured</p>
+                )}
+              </div>
+            </RegionDetail>
+            <RegionDetail title={"Payment providers"}>
+              <div>
+                {region.payment_providers && region.payment_providers.length ? (
```
adrien2p, line 115: suggestion: Should we have an util for that behaviour, something like
```ts
const method = ({
  collection,
  mapperFunction,
  labelProp,
  itemCountToDisplay,
}) => {
  return (
    <>
      {collection?.length ? (
        <div className="flex items-center gap-x-xsmall">
          <p>
            {collection.payment_providers
              .slice(0, itemCountToDisplay)
              .map((p) => mapperFunction(p.id)[labelProp])
              .join(", ")}
          </p>
          {collection.length > 4 && (
            <Tooltip
              side="right"
              content={
                <ul>
                  {collection.slice(itemCountToDisplay).map((p) => {
                    return <li key={p.id}>{mapperFunction(p.id)}</li>
                  })}
                </ul>
              }
            >
              <span className="cursor-default">
                + {collection.length - 4} more
              </span>
            </Tooltip>
          )}
        </div>
      ) : (
        <p>No payment providers configured</p>
      )}
    </>
  )
}
```

and then we can use it like:

```ts
{
    method({
      collection: region.payment_providers,
      mapperFunction: paymentProvidersMapper,
      labelProp: "label",
      itemCountToDisplay: 4
    })
}
```

Of course I did not typed it and the mapper function can be optional and the style need to be managed :)
what do you think? since that behaviour is used multiple times I think it could be nice to centralise and make it generic

adrien2p (commented):
here is a partial review of the settings, some other comments are available from the different testing and can be found [here](https://linear.app/medusajs/issue/CORE-518/domainsettings)

pKorsholm (commented):
discounts lgtm

pKorsholm (commented):
gift cards lgtm 🚀

pKorsholm:
> > > Screenshot 2022-09-07 at 15 49 10
> > > For some reasons, depending on the amount I put in the price input, I get an error 500. With more reasonable amount it works.
> > > Also, we should maybe disallow a price that is greater than the max sub total no?
> > 
> > 
> > Min/Max. subtotal refers to the subtotal of a cart. So these requirements specify that the shipping option is available to carts where the subtotal is >= Min. subtotal and <= Max. subtotal, so it's not related to the price of the shipping option. But interesting with the unknown error on large amounts, but I guess that should be opened as a bug against the core.
> 
> Understood, sorry for the miss-understanding here. You are right for the bug in the core, @olivermrbl do you have the logs otherwise we can try it locally. I'll create the issue in linear

it's because amount is created as an integer in the database. Any amount exceeding 21,474,836.47 will cause an error :smile:

adrien2p:
> > > > Screenshot 2022-09-07 at 15 49 10
> > > > For some reasons, depending on the amount I put in the price input, I get an error 500. With more reasonable amount it works.
> > > > Also, we should maybe disallow a price that is greater than the max sub total no?
> > > 
> > > 
> > > Min/Max. subtotal refers to the subtotal of a cart. So these requirements specify that the shipping option is available to carts where the subtotal is >= Min. subtotal and <= Max. subtotal, so it's not related to the price of the shipping option. But interesting with the unknown error on large amounts, but I guess that should be opened as a bug against the core.
> > 
> > 
> > Understood, sorry for the miss-understanding here. You are right for the bug in the core, @olivermrbl do you have the logs otherwise we can try it locally. I'll create the issue in linear
> 
> it's because the field `ShippingOption.amount` is created as an integer in the database. Any amount exceeding 21,474,836.47 will cause an error 😄

@pKorsholm ahah it was what i thought, but we should still display a proper error message instead of unknown. Wdyt? The validator can take of that and the client can set a max value

src/domain/settings/regions/components/region-form/use-store-data.tsx
kasperkristensen: There is no common pattern between different places where we are creating currency options, some use `currency_code.toUpperCase()` as the label others use the currency name. When we have a single approach to rendering currency and country option then we can do it, but until then I think this is fine as the caching makes sure that the request to fetch store is only done once and shared across everywhere that it is used. 😄

src/domain/pricing/pricing-form/sections/configuration.tsx
```diff
@@ -8,26 +8,27 @@ import TimePicker from "../../../../components/atoms/date-picker/time-picker"
...
 const checkForEnabledConfigs = (
-  config: Record<ConfigurationField, unknown>
+  config: ConfigurationFields
 ): string[] => {
   const enabledConfigs: string[] = []
 
-  if (config.customer_groups?.length > 0) {
-    enabledConfigs.push(ConfigurationField.CUSTOMER_GROUPS)
+  if (config.customer_groups && config.customer_groups.length > 0) {
+    enabledConfigs.push("customer_groups")
```
adrien2p, line 25: question: Why did we move from an enum to strings?

adrien2p: something like that would work?
```ts
const checkForEnabledConfigs = (
  config:  Record<ConfigurationField, unknown>
): string[] => {
  const enabledConfigs: string[] = []

  if (Array.isArray(config.customer_groups) && config.customer_groups.length > 0) {
```

adrien2p (commented):
LGTM pricing: few comments 🎉

src/domain/pricing/details/sections/prices-details/prices-table/use-columns.tsx
```diff
@@ -19,9 +19,7 @@ const usePricesColumns = () => {
                   className="h-full object-cover rounded-soft"
                 />
               ) : (
-                <div className="flex items-center justify-center w-full h-full rounded-soft bg-grey-10">
-                  <ImagePlaceholder size={16} />
-                </div>
+                <ImagePlaceholder />
```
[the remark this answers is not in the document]
pKorsholm: lets do this after the merge 😄

src/domain/pricing/pricing-form/sections/configuration.tsx
```diff
@@ -83,16 +84,17 @@ const Configuration: React.FC<ConfigurationProps> = () => {
...
+                  const ensuredDate = value || new Date()
                   return (
                     <>
                       <DatePicker
-                        date={value}
+                        date={ensuredDate}
                         label="Start date"
                         onSubmitDate={onChange}
                       />
                       <TimePicker
```
pKorsholm, line 96: when selecting dates in the early hours of the morning or late at night the ensuredDate will shift either a day forward or a day back

pKorsholm: It is still possible to set the dates correctly (you just have to go back and set them again), so lets not take this up right now

src/domain/pricing/details/sections/prices-details/prices-table/use-columns.tsx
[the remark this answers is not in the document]
adrien2p: 👍 yeah sure, it was just that I thought it was a broken checkbox at first 😂

src/domain/pricing/pricing-form/sections/configuration.tsx
kasperkristensen: I think that is just how the DatePicker has always been, since I can't see why the new version of RHF or using the little `ensuredDate` hack would result in this shifting.

adrien2p (commented):
LGTM Components/template 👍

src/components/templates/address-form.tsx
```diff
@@ -18,10 +18,16 @@ export type AddressPayload = {
   phone: string | null
 }
 
+export declare enum Type {
+  SHIPPING = "shipping",
+  BILLING = "billing",
+  ADDRESS = "address",
```
adrien2p, line 24: question: This one seams not being used, should we keep it?

kasperkristensen: Wdym? I just added it and its being used in 4 files 😄

adrien2p: I didn't mean the enum, but the address value 😂

src/components/templates/address-form.tsx
```diff
@@ -1,88 +1,67 @@
...
-      }
-    })
-    .filter(Boolean)
-
-  const [selectedCountry, setSelectedCountry] = useState(
-    countryOptions.find((o) => o.value === country)
-  )
-
-  form.register(`${type}.country_code`)
+export type AddressPayload = {
```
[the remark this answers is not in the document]
kasperkristensen: Yes it does, that way you can create a large form such as Order: 

type OrderForm = {
    customer: CustomerPayload
    billing_address: AddressPayload
    shipping_address: AddressPayload
}

const form = useForm<OrderForm>()

And then pass it to this component as <AddressForm form={nestedForm(form, "billing_address")}/>, so having it here and exporting it means that we wont have to create the same form type for address paylaod everywhere that is part of a bigger form.

fPolic (commented):
LGTM: customers ✅ 
LGTM: fundamentals ✅

src/components/templates/collection-modal/index.tsx
```diff
@@ -33,25 +51,81 @@ const CollectionModal: React.FC<CollectionModalProps> = ({
...
+        title: collection.title,
+        handle: collection.handle,
+      })
 
       if (collection.metadata) {
         Object.entries(collection.metadata).map(([key, value]) => {
-          const newMeta = metadata
-          newMeta.push({ key, value })
-          setMetadata(newMeta)
+          if (typeof value === "string") {
```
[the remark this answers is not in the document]
kasperkristensen: The Metadata component only supports editing strings atm, so I just added this to prevent any breaking data types from being loaded in. As an example someone could have metadata: { factory_details: { country: "France", date: "2022-09-03", overseen_by: "Adrien" }} which would break the UI if we try to load it. We need to update the metadata component to handle more, if not all, types of data, but it's out of scope for this PR so I just added this safeguard in the meantime.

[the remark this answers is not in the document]
adrien2p: Alright, got it, just wanted to understand the details since i am not very aware of all details in the admin 😂 sorry for the newby questions

src/components/molecules/select/next-select/components/menu.tsx
```diff
@@ -0,0 +1,322 @@
...
+  cx,
+  children,
+  innerProps,
+  innerRef,
+  placement,
+  selectProps: { onMenuClose, menuIsOpen },
+}: MenuProps<Option, IsMulti, Group>) => {
+  useEffect(() => {
+    const closeOnResize = () => {
+      if (menuIsOpen) {
```
fPolic, line 36: should be a dep

src/components/molecules/select/next-select/components/menu.tsx
```diff
@@ -0,0 +1,322 @@
...
+
+  useEffect(() => {
+    if (
+      document.activeElement !== null &&
+      document.activeElement === ref.current
+    ) {
+      setIsFocused(true)
+    } else {
+      setIsFocused(false)
+    }
```
fPolic, line 134:
```suggestion
    setIsFocused(
          document.activeElement !== null &&
          document.activeElement === ref.current
    )
```

fPolic (commented):
LGTM: molecules ✅

src/domain/settings/return-reasons/create-reason-modal.tsx
```diff
@@ -48,51 +61,50 @@ const CreateReturnReasonModal = ({
...
+                {...register("value", { required: true })}
+                label="Value"
+                placeholder="wrong_size"
+              />
+              <Input
+                className="ml-base"
+                {...register("label", { required: true })}
+                label="Label"
+                placeholder="Wrong size"
+              />
```
pKorsholm, line 77: nit: mark these as required
```suggestion
              <Input
                {...register("value", { required: true })}
                label="Value"
                placeholder="wrong_size"
                required
              />
              <Input
                className="ml-base"
                {...register("label", { required: true })}
                label="Label"
                placeholder="Wrong size"
                required
              />
```

src/domain/settings/taxes/edit-form.tsx
```diff
@@ -76,14 +105,38 @@ const EditTaxRate = ({ modalContext, regionId, taxRate, onDismiss }) => {
...
+            placeholder="12"
+            {...register("rate", { min: 0, max: 100, required: true })}
+            className="mb-base min-w-[335px] w-full"
+          />
+          <Input
+            placeholder="1000"
+            label="Code"
+            {...register("code", { required: true })}
+            className="mb-base min-w-[335px] w-full"
+          />
```
pKorsholm, line 131: Mark as required
```suggestion
          <Input
            label="Name"
            placeholder="Rate name"
            {...register("name", { required: true })}
            className="mb-base min-w-[335px] w-full"
            required
          />
          <Input
            type="number"
            min={0}
            max={100}
            step={0.01}
            label="Rate"
            placeholder="12"
            {...register("rate", { min: 0, max: 100, required: true })}
            className="mb-base min-w-[335px] w-full"
            required
          />
          <Input
            placeholder="1000"
            label="Code"
            {...register("code", { required: true })}
            className="mb-base min-w-[335px] w-full"
            required
          />
```

src/domain/settings/taxes/edit-tax-rate-details.tsx
```diff
@@ -1,34 +1,43 @@
...
   return (
     <div>
       <p className="inter-base-semibold mb-base">Details</p>
       <Input
         disabled={lockName}
-        name="name"
         label="Name"
         placeholder={lockName ? "Default" : "Rate name"}
-        ref={register({ required: !lockName })}
+        {...register("name", { required: !lockName })}
```
pKorsholm, line 24: Mark required
```suggestion
        {...register("name", { required: !lockName })}
        required={!lockName}
```

src/domain/settings/taxes/edit-tax-rate-details.tsx
```diff
@@ -1,34 +1,43 @@
...
       <Input
         type="number"
         min={0}
         max={100}
         step={0.01}
-        name="rate"
         label="Rate"
         placeholder="12"
-        ref={register({ min: 0, max: 100, required: true })}
+        {...register("rate", { min: 0, max: 100, required: true })}
```
pKorsholm, line 34: mark required
```suggestion
        {...register("rate", { min: 0, max: 100, required: true })}
        required
```

src/domain/settings/taxes/edit-tax-rate-details.tsx
```diff
@@ -1,34 +1,43 @@
...
         className="mb-base min-w-[335px] w-full"
       />
       <Input
         placeholder="1000"
-        name="code"
         label="Code"
-        ref={register({ required: true })}
+        {...register("code", { required: true })}
```
pKorsholm, line 40: mark required
```suggestion
        {...register("code", { required: true })}
        required
```

src/domain/settings/taxes/edit.tsx
```diff
@@ -8,6 +8,8 @@ import LayeredModal, {
 import EditForm, { SimpleEditForm } from "./edit-form"
 
 const EditTaxRate = ({ taxRate, taxRateId, regionId, onDismiss }) => {
+  console.log(taxRate)
+
```
pKorsholm, line 12: rm
```suggestion
```

src/domain/settings/taxes/new.tsx
```diff
@@ -84,38 +99,35 @@ const NewTaxRate = ({ regionId, onDismiss }) => {
             <div>
               <p className="inter-base-semibold mb-base">Details</p>
               <Input
-                name="name"
+                {...register("name", { required: true })}
```
pKorsholm, line 102: set required
```suggestion
                {...register("name", { required: true })}
                required
```

src/domain/settings/taxes/new.tsx
```diff
@@ -84,38 +99,35 @@ const NewTaxRate = ({ regionId, onDismiss }) => {
...
-                ref={register({ required: true })}
                 className="mb-base min-w-[335px] w-full"
               />
               <Input
                 type="number"
                 min={0}
                 max={100}
                 step={0.01}
-                name="rate"
+                {...register("rate", { min: 0, max: 100, required: true })}
```
pKorsholm, line 112: set required
```suggestion
                {...register("rate", { min: 0, max: 100, required: true })}
                required
```

src/domain/settings/taxes/new.tsx
```diff
@@ -84,38 +99,35 @@ const NewTaxRate = ({ regionId, onDismiss }) => {
...
                 label="Rate"
                 placeholder="12"
-                ref={register({ min: 0, max: 100, required: true })}
                 className="mb-base min-w-[335px] w-full"
               />
               <Input
                 placeholder="1000"
-                name="code"
+                {...register("code", { required: true })}
```
pKorsholm, line 119:
```suggestion
                {...register("code", { required: true })}
                required
```

pKorsholm (commented):
settings look good, we are missing some fields being marked as required, but nothing major.

Also we still don't have support in admin for nested return reasons, (this shouldn't be handled here)

pKorsholm (commented):
fundamentals lgtm

src/components/organisms/edit-user-modal/index.tsx
```diff
@@ -1,93 +1,101 @@
...
+        <Modal.Body>
+          <Modal.Header handleClose={handleClose}>
+            <span className="inter-xlarge-semibold">Edit User</span>
+          </Modal.Header>
+          <Modal.Content>
+            <div className="w-full flex gap-x-base mb-base">
+              <InputField
+                label="First Name"
+                placeholder="First name..."
+                {...register("first_name", { required: true })}
```
adrien2p, line 62: question: In some places there is both required on the input and in the register, in other places it is only set in the register. should we make the input required too?

adrien2p (commented):
LGTM organism

src/components/organisms/accordion/index.tsx
```diff
@@ -101,7 +101,7 @@ Accordion.Item = Item
 
 const MorphingTrigger = () => {
   return (
-    <div className="p-[6px] relative group btn-ghost">
+    <div className="p-[6px] relative group btn-ghost rounded-rounded">
```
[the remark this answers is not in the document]
kasperkristensen: It's in the tailwind.config.js `borderRadius: { ..., rounded: "8px", ....}`

src/domain/orders/new/new-order.tsx
```diff
@@ -1,311 +1,130 @@
...
+            option_id: data.shipping_option.value,
+            price: data.custom_shipping_price
+              ? persistedPrice(
+                  region?.currency_code!,
+                  data.custom_shipping_price
+                )
+              : undefined,
+          },
+        ],
+        shipping_address: data.shipping_address_id
```
adrien2p, line 65: suggestion: the API expect an address here, but if there is an if then we assign the id (string). should we remove the first ternary part?

src/domain/orders/new/new-order.tsx
```diff
@@ -1,311 +1,130 @@
...
+              country_code: data.shipping_address?.country_code?.value,
+              company: data.shipping_address?.company || undefined,
+              first_name: data.shipping_address?.first_name,
+              last_name: data.shipping_address?.last_name,
+              phone: data.shipping_address?.phone || undefined,
+              postal_code: data.shipping_address?.postal_code,
+              province: data.shipping_address?.province || undefined,
+            }
+          : undefined,
+        billing_address: data.billing_address_id
```
adrien2p, line 81: suggestion: same as previous

adrien2p (commented):
Very nice Order :)

pKorsholm (commented):
lgtm template 
lgtm organisms

src/domain/orders/details/address-modal.tsx
```diff
@@ -1,157 +1,153 @@
...
+  AddressPayload,
+  Type,
+} from "../../../components/templates/address-form"
+import useNotification from "../../../hooks/use-notification"
+import { isoAlpha2Countries } from "../../../utils/countries"
+import { getErrorMessage } from "../../../utils/error-messages"
+import { nestedForm } from "../../../utils/nested-form"
+
+type AddressType =
+  | AdminPostOrdersOrderReq["shipping_address"]
```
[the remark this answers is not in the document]
kasperkristensen: There is definetly a better approach to it than this, but your suggestions won't align with the types of update mutations that is passed into this modal. Will leave it as is for now, but definetly something to revisit!

src/domain/gift-cards/manage/sections/images.tsx
```diff
@@ -1,106 +1,100 @@
...
-          render={({ value, onChange }) => {
+          control={form.control}
+          render={({ field: { value, onChange } }) => {
             return (
               <RadioGroup.Root
-                value={value}
+                value={value ? `${value}` : undefined}
                 onValueChange={(value) => {
-                  onChange(value)
+                  onChange(parseInt(value))
```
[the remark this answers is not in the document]
kasperkristensen: Yea let us leave it for now, I could imagine that we are going to add the same principal of seperating images and thumbnail like we've done for products in the near future :D

src/components/organisms/accordion/index.tsx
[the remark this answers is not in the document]
kasperkristensen: Yea rounded is Tailwinds shorthand for border radius, and rounded is our class name for `8px`, might have to come up with some better names at some point haha

pKorsholm (commented):
molecules lgtm

src/domain/orders/new/new-order.tsx
```diff
@@ -1,311 +1,130 @@
...
+              ? persistedPrice(
+                  region?.currency_code!,
+                  data.custom_shipping_price
+                )
+              : undefined,
+          },
+        ],
+        shipping_address: data.shipping_address_id
+          ? data.shipping_address_id
+          : !isNullishObject(data.shipping_address)
```
kasperkristensen, line 67: This is a bit tricky because the API typing is wrong, it should accept either a id or a address. Also there is currently a bug i nthe core, that prevents the billing address to be added if it's not added as an id. I have an open PR to fix, that I need to get up to date an have merged.

src/components/organisms/invite-modal/index.tsx
```diff
@@ -1,80 +1,109 @@
...
+                  return (
+                    <Select
+                      label="Role"
+                      onChange={onChange}
+                      options={roleOptions}
+                      value={value}
+                    />
+                  )
+                }}
+              />
```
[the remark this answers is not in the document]
kasperkristensen: At the same time, some user might be using the role for something, so I would argue that we should display it even if "vanilla" Medusa does not use it for anything

src/domain/orders/new/new-order.tsx
kasperkristensen: This is a bit tricky because the API typing is wrong, it should accept either a id or a address. Also there is currently a bug i nthe core, that prevents the billing address to be added if it's not added as an id. I have an open PR to fix, that I need to get up to date an have merged.

adrien2p (commented):
LGTM atoms 👍

fPolic (commented):
LGTM: discounts ✅

src/domain/gift-cards/manage/sections/information.tsx
```diff
@@ -104,30 +109,22 @@ const Information: React.FC<InformationProps> = ({ giftCard }) => {
...
-              label="Subtitle"
-              name="subtitle"
-              placeholder="Add a subtitle"
-              defaultValue={giftCard?.subtitle}
-              ref={register}
-            />
-          </div>
+        <div className="grid grid-cols-2 gap-large">
           <Input
+            label="Name"
```
adrien2p, line 114: question: should the field be ;marked as required as well? (general question) as already mentioned else where 😂

adrien2p (commented):
very nice 👍 gift-cards

src/domain/gift-cards/details/index.tsx
```diff
@@ -25,12 +26,12 @@ import UpdateBalanceModal from "./update-balance-modal"
...
   id,
 }) => {
-  const { gift_card, isLoading } = useAdminGiftCard(id!, {
+  const { gift_card: giftCard, isLoading } = useAdminGiftCard(id!, {
     enabled: !!id,
   })
   const { regions } = useAdminRegions()
 
-  const updateGiftCard = useAdminUpdateGiftCard(gift_card?.id!)
+  const updateGiftCard = useAdminUpdateGiftCard(giftCard?.id!)
```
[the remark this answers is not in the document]
kasperkristensen: Would cause an error if gift_card is not load loaded yet and we then try to access id. Doing it like giftCard?.id! makes sense as we avoid errors when the page is first loaded, but at the time where updateGiftCard is actually invoked we know that the gift card is loaded.

src/domain/gift-cards/details/index.tsx
```diff
@@ -59,7 +60,7 @@ const GiftCardDetails: React.FC<RouteComponentProps<{ id: string }>> = ({
     },
   ]
 
-  const handleUpdate = (data) => {
+  const handleUpdate = (data: AdminPostGiftCardsGiftCardReq) => {
     updateGiftCard.mutate(
       { ...data },
```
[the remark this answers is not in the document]
kasperkristensen: The wrong type is used in medusa-react, AdminPostGiftCardsGiftCardReq is the correct type of payload since it is an update.

src/domain/gift-cards/manage/form/mappers.ts
```diff
@@ -1,45 +1,55 @@
...
-  if (values.tags) {
-    payload["tags"] = values.tags.map((tag) => ({ value: tag }))
+  if (values.thumbnail) {
+    imagePayload.thumbnail = imagePayload.images?.length
+      ? imagePayload.images[values.thumbnail]
+      : undefined
   }
 
-  return payload
+  return {
```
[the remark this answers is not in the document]
kasperkristensen: Hmm good point, but if we set it to null would that not result in us removing the gift card from it's sales channel, thus making it unavailable? Also think that the API should accept undefined for this as you might not want to update SC's and again it is annoying having to send existing data to the update endpoint.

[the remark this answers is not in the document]
kasperkristensen: Will leave it as is, as I think undefined is the intended payload for this update, as we don't want to update SC's.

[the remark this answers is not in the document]
adrien2p: Agree, the api should allow undefined in the update end point 👍 new ticket to add 😂

src/domain/gift-cards/manage/form/gift-card-form-context.tsx
```diff
@@ -1,96 +1,87 @@
...
-  }
-
-  const removeImage = (image) => {
-    setHasImagesChanged(true)
-    const tmp = images.filter((img) => img.url !== image.url)
-    setImages(tmp)
-  }
-
-  const methods = useForm()
+  const methods = useForm<MangeGiftCardFormData>()
```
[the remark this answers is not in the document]
kasperkristensen: Haha, should be good now 😅

olivermrbl (approved):

PR #683, opened 2022-09-06
feat: Revamp region settings

kasperkristensen:
**What**
- Revamps region settings to use the new version of RHF
- Revamps design of region settings


src/domain/settings/regions/new/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+                <RegionDetailsForm
+                  form={nestedForm(form, "details")}
+                  isCreate
+                />
+              </Accordion.Item>
+              <Accordion.Item
+                title="Providers"
+                value="providers"
+                forceMountContent
+              >
```
pKorsholm, line 131: Question: should we mark this section as required since it's collapsed and the fields inside are required?

src/domain/settings/regions/index.tsx
```diff
@@ -1,138 +1,27 @@
...
-  const handleSelect = (id: string) => {
-    refetch().then(() => {
-      setSelectedRegion(id)
-      document.getElementById(id)?.scrollIntoView({
-        behavior: "smooth",
-      })
-    })
-  }
+const RegionsRoute = (props: Props) => {
+  const filepath: string | undefined = props["*"]
```
pKorsholm, line 10: nit: naming

pKorsholm (commented):
Very nice! 😍  just have a couple of comments regarding required fields and a nit!

PR #686, opened 2022-09-12
Feat(admin): Add copy invite link in teams page

pKorsholm:
**What**
- Add "copy invite link" to actionables for invites in the user table
- Update format of invite link template in store settings

**How**
- lift admin url from store, if it's not present on the store use the current window url as a base for generating the invite link using `window.location.origin`

Fixes CORE-453


srindom (approved):
LGTM!

PR #688, opened 2022-09-14
feat: tax inclusive pricing in region settings

kasperkristensen:
**What**

- Adds toggle to decide if region has TIP.
- Added new TIP Tooltip.
- Show whether region has TIP when adding shipping prices and requirements.

>> Note might need to move the TIP toggle to tax settings depending on this discussion: https://www.figma.com/file/PluqU9Ind9yvD9JbGCkI35?node-id=3231:11131#265755434 <<

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
   },
   "dependencies": {
     "@hookform/error-message": "^2.0.0",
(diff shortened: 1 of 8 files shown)


fPolic (approved):
LGTM 👍

PR #689, opened 2022-09-14
feat: add TIP indicator to variant prices

kasperkristensen:
**What**
- Adds an indicator to show if a variant price is a TIP.


src/domain/products/components/prices-form/index.tsx
```diff
@@ -63,6 +64,16 @@ const PricesForm = ({ form }: Props) => {
...
+        })
+      }
+
+      const index = fields.findIndex((field) => field.region_id === reg.id)
+
+      if (index !== -1 && fields[index].includes_tax !== reg.includes_tax) {
+        update(index, {
+          ...fields[index],
+          includes_tax: reg.includes_tax,
         })
```
olivermrbl, line 77: suggestion(non-blocking): Change naming both of these checks (region setting and currency setting) to avoid ambiguity

olivermrbl: Could be something like `const regionTaxSetting` and `const currencyTaxSetting`

olivermrbl (approved):
LGTM, just a single suggestion 👍

PR #690, opened 2022-09-14
feat: revamp currencies settings

kasperkristensen:
**What**
- Revamps the design and functionality of currency settings.
- Fixes bug where NextSelect was not re-rendered when options were changed.

**Note**
Searching in currency tables is missing from the implementation as it's not supported by the core atm.

Closes CORE-401


src/domain/settings/currencies/components/currency-tax-setting.tsx
```diff
@@ -0,0 +1,85 @@
...
+      includes_tax: currency.includes_tax,
+    })
+  }, [currency])
+
+  const onSubmit = handleSubmit((data: CurrencyTaxSettingFormType) => {
+    mutate(data, {
+      onSuccess: () => {
+        notification("Success", "Successfully updated currency", "success")
+      },
+      onError: (error) => {
```
fPolic, line 40: suggestion: using `mutateAsync` will properly catch error, this way if we have 4xx the onError won't fire

src/domain/settings/currencies/components/currency-tax-setting.tsx
```diff
@@ -0,0 +1,85 @@
...
+        <Controller
+          control={control}
+          name="includes_tax"
+          render={({ field: { value, onChange } }) => {
+            return (
+              <Switch
+                checked={value}
+                onCheckedChange={(data) => {
+                  onChange(data)
+                  onSubmit()
```
fPolic, line 74: also if there is an error on submit, we should revert toggle state

src/domain/settings/currencies/components/store-currencies/add-currencies-screen.tsx
```diff
@@ -0,0 +1,153 @@
...
+    useRowSelect
+  )
+
+  if (status === "error") {
+    return <div>Failed to load</div>
+  }
+
+  if (status === "loading" && !currencies) {
+    return <div>Loading...</div>
+  }
```
fPolic, line 91: suggestion: can we add a bit of spacing around these messages
Screenshot 2022-09-15 at 10 58 35

fPolic (commented):
Really nice work @kasperkristensen! I've added a couple of minor comments

src/domain/settings/currencies/components/currency-tax-setting.tsx
kasperkristensen: Why would it not catch the errors as is? Just tried it, and catches 400s without issue, is this not what you are experiencing?

src/domain/settings/currencies/components/currency-tax-setting.tsx
kasperkristensen: Definitely, I will add it asap 💪

src/domain/settings/currencies/components/currency-tax-setting.tsx
fPolic: Hmm, I get an uncaught error if it's 404

kasperkristensen: Okay super interesting, can you give me some steps to try and reproduce? If I ensure a 404, it catches it properly and gives the correct error message in a notification.

image

olivermrbl (changes requested):
This is great work @kasperkristensen - really neat user experience!

todo: TIP components should be hidden behind our feature flag component

src/domain/settings/currencies/components/currency-tax-setting.tsx
kasperkristensen: Okay, did some digging and it seems like `medusa-react` can't handle 404's that a thrown as a result of a route not existing, but handles "actual" 404's perfectly fine.

PR #692, opened 2022-09-15
fix: create product with sales channels disabled

kasperkristensen:
**What**

- Hides all functionality related to sales channels behind feature flags to prevent issues with using the product domain without sales channels enabled.


olivermrbl (approved):
LGTM!

PR #696, opened 2022-09-16
fix: batch job descriptions

fPolic:
**What**
- extend batch job descriptions to support other states
- reset batch job id when import file is removed

### DIFF
--- a/src/components/organisms/batch-jobs-activity-list/index.tsx
+++ b/src/components/organisms/batch-jobs-activity-list/index.tsx
@@ -122,16 +122,17 @@ const BatchJobActivityCard = ({ batchJob }: { batchJob: BatchJob }) => {
   const getBatchJobFileCard = () => {
     const twentyfourHoursInMs = 24 * 60 * 60 * 1000
 
-    const iconColor =
-      Math.abs(relativeTimeElapsed.raw) > twentyfourHoursInMs
-        ? "#9CA3AF"
-        : undefined
-
     const icon =
       batchJob.status !== "completed" && batchJob.status !== "canceled" ? (
         <Spinner size={"medium"} variant={"secondary"} />
       ) : (
-        <FileIcon fill={iconColor} size={20} />
+        <FileIcon
+          className={clsx({
+            "text-grey-40":
+              Math.abs(relativeTimeElapsed.raw) > twentyfourHoursInMs,
+          })}
+          size={20}
+        />
       )
 
     const fileSize = batchJob.result?.file_key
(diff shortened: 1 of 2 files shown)


src/components/organisms/batch-jobs-activity-list/index.tsx
```diff
@@ -134,12 +134,16 @@ const BatchJobActivityCard = ({ batchJob }: { batchJob: BatchJob }) => {
         <FileIcon fill={iconColor} size={20} />
```
kasperkristensen, line 134:
```suggestion
        const iconColor = clsx({ "text-grey-40": Math.abs(relativeTimeElapsed.raw) > twentyfourHoursInMs })
       
       // ...
       
        <FileIcon className={iconColor} size={20} />
```

nit: We should avoid hardcoding color values, as it makes it more difficult in the case of updates to our design system.

src/domain/products/batch-job/import.tsx
```diff
@@ -141,6 +141,7 @@ function ImportProducts(props: ImportProductsProps) {
         deleteFile({ file_key: fileKey })
       }
       cancelBathJob()
+      setBatchJobId(undefined)
     } catch (e) {
       console.log(e)
```
kasperkristensen, line 146: nit: Can we do some better error handling in this case? Generally don't think that we should be outputing errors to the console, at least not in production.

kasperkristensen (approved):
LGTM with some minor nits

src/domain/products/batch-job/import.tsx
fPolic: agree - I removed the log and added notifications in case requests fail

src/components/organisms/batch-jobs-activity-list/index.tsx
```diff
@@ -122,24 +122,29 @@ const BatchJobActivityCard = ({ batchJob }: { batchJob: BatchJob }) => {
...
+        />
       )
 
-    const fileSize =
-      batchJob.status !== "canceled"
-        ? batchJob.result?.file_key
-          ? bytesConverter(batchJob.result?.file_size ?? 0)
-          : `Preparing ${operation.toLowerCase()}...`
-        : undefined
+    const fileSize = batchJob.result?.file_key
```
olivermrbl, line 138: praise: Nice!

olivermrbl (approved):
LGTM!

olivermrbl:
@kasperkristensen Requested changes have been added, so I'll go ahead and merge 👍

PR #697, opened 2022-09-16
feat: adds price list import

srindom:
Relies on PriceList import strategy available with core PR: https://github.com/medusajs/medusa/pull/2210

https://user-images.githubusercontent.com/7554214/190674294-1569ef86-7eec-486d-bef4-dc07a4ecfd59.mov

Fixes CORE-591


src/domain/pricing/batch-job/import.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+  /**
+   * Returns create/update counts from stat descriptor.
+   */
+  const getSummary = (): { toCreate: number } | undefined => {
+    if (!batchJob) {
+      return undefined
+    }
+
+    const res = batchJob.result?.stat_descriptors![0].message.match(/\d+/g)
```
fPolic, line 124:
```suggestion
    const res = batchJob.result?.stat_descriptors?[0].message.match(/\d+/g)
```

fPolic (commented):
Nice work! We should add the new batch job type [here](https://github.com/medusajs/admin/blob/develop/src/components/organisms/batch-jobs-activity-list/index.tsx#L51:L54) so the activity drawer works with the new type

fPolic (approved):
LGTM 🎉

PR #698, opened 2022-09-19
fix: amount inputs work with locales that use comma separators

### LINKED ISSUE #630: Can't add price with cents/fractions
Steps to reproduce

1. Edit price of a variant
2. Type in fractional price, in example 19,50
3. After saving see the price rounded down to 19

Price is saved in database as 1900, should be 1950.

kasperkristensen:
**What**

- Fixes a bug where cents/fractions were ignored when using the admin dashboard in a browser with a locale that uses commas as the decimal separator.

**How**
- Instead of using `parseFloat` to get the value that should be used to calculate the persisted amount, we instead use the float that is returned as part of the `onChange` function in react-currency-input-field.

Closes #630 

Closes CORE-595

### DIFF
--- a/src/components/organisms/currency-input/index.tsx
+++ b/src/components/organisms/currency-input/index.tsx
@@ -10,7 +10,7 @@ import React, {
 import AmountField from "react-currency-input-field"
 import { Option } from "../../../types/shared"
 import { currencies, CurrencyType } from "../../../utils/currencies"
-import { getDecimalDigits, normalizeAmount } from "../../../utils/prices"
+import { normalizeAmount, persistedPrice } from "../../../utils/prices"
 import InputError from "../../atoms/input-error"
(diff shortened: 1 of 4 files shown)


src/components/organisms/currency-input/index.tsx
```diff
@@ -189,48 +189,78 @@ const Amount = forwardRef<HTMLInputElement, AmountInputProps>(
...
+    //     value = 0
+    //   }
+
+    //   if (currencyInfo) {
+    //     const amount = parseFloat(value)
+    //     const multiplier = getDecimalDigits(currencyInfo.code)
+    //     persistedAmount = multiplier * amount
+    //   }
+
+    //   if (onChange && typeof persistedAmount !== "undefined") {
```
srindom, line 209: todo: remove

srindom (approved):
LGTM!

PR #699, opened 2022-09-19
fix: realign tax rate inputs across region and tax settings

kasperkristensen:
**What**
- Ensure a uniform experience when adding tax rates in both region and tax settings.

Closes CORE-559

### DIFF
--- a/src/domain/settings/regions/components/region-form/region-details-form.tsx
+++ b/src/domain/settings/regions/components/region-form/region-details-form.tsx
@@ -37,7 +37,7 @@ const RegionDetailsForm = ({ form, isCreate = false }: Props) => {
       <div className="grid grid-cols-2 gap-large">
         <InputField
           label="Title"
-          placeholder="Europe..."
+          placeholder="Europe"
           required
           {...register(path("name"), {
             required: "Title is required",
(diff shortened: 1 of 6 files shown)


olivermrbl (changes requested):
A couple of comments:
- I can't seem to add tax rates with decimals
- Tax rate and tax code are not editable in "Edit Region Details" modal
- Tax rate and tax code should be renamed to "Default Tax Rate" and "Default Tax Code" in regions setting - both create and update

kasperkristensen:
> A couple of comments:
> 
> * I can't seem to add tax rates with decimals
> * Tax rate and tax code are not editable in "Edit Region Details" modal
> * Tax rate and tax code should be renamed to "Default Tax Rate" and "Default Tax Code" in regions setting - both create and update

Addressed feedback, with the exception of `Tax rate and tax code are not editable in "Edit Region Details" modal`, as per Seb this should only be edited from the tax settings.

olivermrbl (approved):
This is great! 💪 

Noticed a minor thing. Countries in "Create Region" flow should be required. If left out, page breaks when trying to map through undefined countries.

Feel free to merge afterwards!

PR #701, opened 2022-09-20
add price type to shipping option modal

pKorsholm:


src/domain/settings/regions/components/shipping-option-form/index.tsx
```diff
@@ -41,6 +49,8 @@ const ShippingOptionForm = ({ form, region, isEdit = false }: Props) => {
     formState: { errors },
   } = form
 
+  const watchPriceType = form.watch("price_type")
```
kasperkristensen, line 52:
```suggestion
  const watchPriceType = useWatch({
    control,
    name: "price_type"
  })
```

Prefer using `useWatch` over `watch`, as it isolates re-renders to the hook rather than re-rendering the entire form when the value changes.

PR #703, opened 2022-09-20
hotfix: avoid duplicate select options

### LINKED ISSUE #700: Select component displays selected elements twice
`NextSelect` and `Select` displaying selected element twice:
Знімок екрана з 2022-09-19 21-28-42
To fix it change in `useSelectProps` hook `unselectedOptions` to:
` const unselectedOptions = stateOptions.filter(
      (option) => !tmp.find((op) => op === option || op.value === option?.value)
    )
`

kasperkristensen:
**What**
- Prevent having duplicate select options in `isMulti` NextSelects

**How**
- use deep equality check to filter unselected options.

Closes #700


olivermrbl (approved):
LGTM!

PR #707, opened 2022-09-21
Add labels for stripe-blik and stripe-przelewy24 payments

chemicalkosek:
This PR adds missing labels for payment providers.


olivermrbl (approved):
Thanks for the contribution @chemicalkosek!

PR #711, opened 2022-09-28
hotfix: prevent error on no regions in store

kasperkristensen:
**What**
- Adds check that prevents error if navigating to `settings/regions` when there are no regions in the store.


olivermrbl (approved):
LGTM!

PR #717, opened 2022-09-29
feat: revamp amount input

kasperkristensen:
**What**
- Implements new AmountInput and new AmountAndCurrencyInput components, to be used for future input fields where an admin enters prices. 

**How**
- `AmountInput` accepts a currencyCode, uses that to return a DB-friendly integer value, and displays a formatted amount.
- `AmountAndCurrencyInput` allows the user to make changes to both the currency and amount of a price from one component.

**AmountAndCurrencyInput**
image
image
image

**AmountInput**
image

Closes CORE-407


srindom (approved):
LGTM - soo nice components! 🤤

PR #724, opened 2022-10-10
feat: "Edit order" modal

fPolic:
**What**
- implemented order edit modal on the order details page

**TODOs** 
- [x] check how we want to display prices in the "Add variant" table
- [x] fix for order relations
- [x] fix cannot delete an edit if changes were made (BD issue due to row reference) 
   - https://github.com/medusajs/medusa/pull/2420
- [x] delete order edit when focus modal is closed with outside click

--- 

RESOLVES CORE-485


srindom:
@fPolic - what are next steps on this one? Can see you have a couple of todos listed - are you planning on picking this up or how do you suggest we proceed?

fPolic:
> @fPolic - what are next steps on this one? Can see you have a couple of todos listed - are you planning on picking this up or how do you suggest we proceed?

Yes I'm working on the last one rn and need to determine how do we approach the first one for this to be done. For other TODOs I was thinking to open a ticket since those issues are not directly related to the scope of the ticket, wdyt?

srindom:
Remove tooltip: 

CleanShot 2022-10-11 at 4 34 49

srindom:
> Yes I'm working on the last one rn and need to determine how do we approach the first one for this to be done. For other TODOs I was thinking to open a ticket since those issues are not directly related to the scope of the ticket, wdyt?

Sounds good - a quick note on the `order.edits` relation though.

I think it, would be great if we could fetch these separately by calling `GET /admin/order-edits?order_id={id}` instead - this would also be in line with moving away from DB joins. It is non-blocking, but will be a more sustainable solution going forward.

srindom:
When "new total" > "previous total" the difference due should be green and show as positive amount; in the opposite case 
it should be red and show as a negative. I.e., reverse the current implementation logic here


CleanShot 2022-10-11 at 4 42 10

src/domain/orders/edit/modal.tsx
```diff
@@ -0,0 +1,391 @@
...
+  let displayItems = orderEdit.items.sort(
+    // @ts-ignore
+    (a, b) => new Date(a.created_at) - new Date(b.created_at)
+  )
+
+  if (filterTerm) {
+    displayItems = displayItems.filter((i) =>
+      i.title.toLowerCase().includes(filterTerm)
+    )
+  }
```
srindom, line 215: todo: Make it possible to filter by `i.variant.sku` too - if present

srindom:
Regarding displaying prices - I know that @kasperkristensen did a similar exercise when revamping the RMA flows. Basically, we should include prices with the Region and Customer context as seen here: https://github.com/medusajs/admin/blob/3fead7469181344883b985f6bd8d61d4ad33b51f/src/domain/orders/components/items-to-send-form/add-additional-items-screen/index.tsx#L56

fPolic:
> Sounds good - a quick note on the `order.edits` relation though.
> 
> I think it, would be great if we could fetch these separately by calling `GET /admin/order-edits?order_id={id}` instead - this would also be in line with moving away from DB joins. It is non-blocking, but will be a more sustainable solution going forward.

Created an issue (CORE-697) since it's now blocked by OE list endpoint.

> Regarding displaying prices - I know that @kasperkristensen did a similar exercise when revamping the RMA flows. 
> Basically, we should include prices with the Region and Customer context as seen here:

Perfect 🙏 done. I fixed the things you pointed out and resolved TODOs. Should be good for review now.

srindom:
When I click **Save and close** - the modal doesn't close. Do you have the same experience on your end?

srindom:
Have issues with the following steps:
1. Create Order Edit
2. Add changes
3. Click **Cancel**, **X**, or outside of the modal

**Expected**
Modal closes edit is deleted

**Actual**
Modal doesn't close; order edit seems to be deleted

In the server log it looks like the order edit is deleted and then immediately a get is attempted
CleanShot 2022-10-13 at 6 23 25

fPolic:
> When I click **Save and close** - the modal doesn't close. Do you have the same experience on your end?

Sorry, the last commit broke it, it's fixed now

srindom:
Store operators should be allowed to add additional items despite them being out of stock; e.g. if they know they are coming back in stock in a few days. 

I.e. there should not be any greyed out items in this list:
CleanShot 2022-10-17 at 11 15 42

srindom:
Also show the price in the view with taxes :)

srindom:
Btw, I am happy to merge this PR as is and then have a separate batch of issues to complete to clean up the rest. Wdyt?

srindom (approved):

PR #725, opened 2022-10-11
refactor: BatchJob polling 

fPolic:
**What**
- refactor batch job polling to minimise the number of requests

**How**
- refactor the polling context to use dynamic increasing intervals for refresh
- retrieve a batch job by using a hook which refreshes automatically depending on the status of the batch job

---

**RESOLVES CORE-648**

### DIFF
--- a/src/components/organisms/activity-drawer/index.tsx
+++ b/src/components/organisms/activity-drawer/index.tsx
@@ -8,11 +8,11 @@ import BatchJobActivityList from "../batch-jobs-activity-list"
 
 const ActivityDrawer = ({ onDismiss }) => {
   const ref = React.useRef<HTMLDivElement>(null)
-  const { batchJobs, hasPollingError, refetchJobs } = useContext(PollingContext)
+  const { batchJobs, hasPollingError, refetch } = useContext(PollingContext)
   useOutsideClick(onDismiss, ref)
 
   useEffect(() => {
-    refetchJobs()
+    refetch()
   }, [])
 
   return (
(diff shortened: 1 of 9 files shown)


olivermrbl (commented):
This is really neat @fPolic 💪 

**A couple of observations:**

I am experiencing the following intervals 5sec, 10sec, 15sec, 90sec. See attached image. The 90sec should be 30sec no?
CleanShot 2022-10-13 at 11 03 58

The timer is not reset upon creating a new BatchJob.

The timer is not reset upon opening the BatchJob drawer.

The latter is debatable, but I believe we should reset the timer every time a new BatchJob is created to ensure more frequent updates right after no?

fPolic:
>  The 90sec should be 30sec no?

@oli - did you leave the page while testing? In that case react-query pauses the interval fetches... I think that this is desired behaviour in this case but we could additionally set `refetchOnWindowFocus` wdyt?

> The timer is not reset upon creating a new BatchJob.

Added 👍

olivermrbl:
Got it makes sense!
> we could additionally set refetchOnWindowFocus wdyt?

I think this is an excellent addition 💪

olivermrbl (approved):
LGTM! Feel free to merge when refetch on window focus has been incorporated 👍

kasperkristensen (approved):
LGMT! Agreed that refetchOnWindowFocus would be great final addition 💪

adrien2p (approved):
Looks pretty awesome 😎

PR #726, opened 2022-10-11
Fix(medusa): prevent showing protected routes when loading unauthenticated admin

pKorsholm:
**What**
- prevent showing the layout in a flash before the login screen when navigating unauthenticated to a page under `/a/*`

**How**
- wrap the components in `pages/a` in a `privateRoute` 

Fixes CORE-439


srindom (approved):
LGTM! Great to finally get this fixed 😅

PR #727, opened 2022-10-11
hotfix: Allow variant price to be 0

kasperkristensen:
**What**

- Allows setting variant prices to 0.
- Brings in the changes from https://github.com/medusajs/admin/pull/698 as this is also part of allowing setting prices to 0.

**Also**
- Updates invalid query params from product table so the `list` products request does not fail.
- Fixes stroke props on tax icon


olivermrbl (approved):
LGTM!

PR #732, opened 2022-10-13
hotfix: Title in shopping options modal

patrikkarlsson:
Misleading title in modal when adding shipping option.


olivermrbl (approved):
Nice catch @patrikkarlsson! Thanks for contribution 💪

PR #735, opened 2022-10-14
hotfix: fix unset prices on variant creation

kasperkristensen:
**What**
- The `createAddPayload` was defaulting unset prices to 0. This was an issue since we now load every possible price that the user can set for a variant. The expected behaviour would be that if the user does not set one of these prices to anything then they won't be added to the `variant.prices` array, but this was not the case. This fix ensures that all prices where the amount is not specifically set (meaning that they are null or undefined) won't be passed on to the server.


olivermrbl (approved):
LGTM

PR #737, opened 2022-10-18
fix: OE UI polish

fPolic:
**What**
- fix bugs, align implementations of the modal and the timeline, fetch optimisations

**TODO**
- [x] use the list endpoint instead of `edits`relation (FIXES CORE-697)
- [x] hide timeline event while editing and handle stale edits (FIXES CORE-688)
- [x] refund OE doesn't show refund time line event for refunding (FIXES CORE-692)
- [x] OE with removed line items shows incorrectly (FIXES CORE-690)
- [x] add spacing in order edit time line event (FIXES CORE-708)
- [x] "Customer payment required" should not be shown if difference due is 0 (FIXES CORE-707)
- [x] copy SKU in edit modal (FIXES CORE-709)
- [x] Timeline event with "by" field should show email when no name is available (FIXES CORE-691)
- [x] show prices with taxes in modal
- [x] allow operators to select variants despite being out of stock
- [x] confirmation-request sent should show based on Notifications (FIXES CORE-718)
- [x] "Spamming" the quantity ticker on order editing results in error (FIXES CORE-717)

### DIFF
--- a/src/components/molecules/timeline-events/order-edit/confirmed-difference-due.tsx
+++ b/src/components/molecules/timeline-events/order-edit/confirmed-difference-due.tsx
@@ -18,13 +18,16 @@ const EditConfirmedDifferenceDue: React.FC<RequestedProps> = ({ event }) => {
   })
 
   const [showRefundModal, setShowRefundModal] = useState(false)
+
   if (
     !event.isLastConfirmed ||
     !order ||
-    (order && order.total >= order.paid_total)
+    (order && order.paid_total - order.refunded_total - order.total > 0)
   ) {
     return null
   }
+
+  const refundableAmount = order.paid_total - order.refunded_total - order.total
 
   return (
     <>
@@ -34,14 +37,6 @@ const EditConfirmedDifferenceDue: React.FC<RequestedProps> = ({ event }) => {
         iconColor={EventIconColor.RED}
         time={event.time}
         isFirst={event.first}
-        // midNode={
-        //   <span className="inter-small-regular text-grey-50">
-        //     {formatAmountWithSymbol({
-        //       amount: orderEdit.difference_due,
-        //       currency: event.currency_code,
-        //     })}
-        //   </span>
-        // }
       >
         <Button
           onClick={() => setShowRefundModal(true)}
@@ -51,7 +46,7 @@ const EditConfirmedDifferenceDue: React.FC<RequestedProps> = ({ event }) => {
         >
           Refund
           {formatAmountWithSymbol({
-            amount: Math.abs(order?.total - order?.paid_total),
+            amount: refundableAmount,
             currency: event.currency_code,
           })}
         </Button>
@@ -59,7 +54,7 @@ const EditConfirmedDifferenceDue: React.FC<RequestedProps> = ({ event }) => {
       {showRefundModal && (
         <CreateRefundModal
           order={order}
-          initialAmount={order.paid_total - order.total}
+          initialAmount={refundableAmount}
           initialReason="other"
           onDismiss={() => setShowRefundModal(false)}
         />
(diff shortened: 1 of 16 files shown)


srindom:
When I filter in the variants table it doesn't show the changes as I type; the table seems to update though

srindom:
thought(non-blocking,if-minor): maybe we should include quantities when we remove/modify items e.g. if you reduce a LineItem right now it will be listed in the timeline as "removed". Wdyt?

srindom:
Really great work @fPolic !! 💪

srindom:
We should update the tooltip when a line item has `isLocked === true` to say:

This line item is part of a fulfillment and cannot be edited. Cancel the fulfillment to edit the line item.

src/components/molecules/timeline-events/order-edit/confirmed-difference-due.tsx
```diff
@@ -1,50 +1,70 @@
-import { useAdminOrderEdit } from "medusa-react"
-import React from "react"
+import { useAdminOrder, useAdminOrderEdit } from "medusa-react"
```
srindom, line 1: general thought on this one: 

I think we should consider showing the refund timeline event regardless of any OrderEdits. The approach would be this:

There is always a check for `order.paid_total > order.total` if the check is >0 then we show the refund timeline event. 

The narrative here is: 
We don't care how this discrepancy between paid and actual total arose, but we owe the customer money so now we should issue a refund. 

I will move this thought to an issue that we can consider separately, justed wanted to note it here.

src/components/molecules/timeline-events/order-edit/requested.tsx
```diff
@@ -1,23 +1,61 @@
...
 const EditRequested: React.FC<RequestedProps> = ({ event }) => {
+  const [showResend, setShowResend] = useState(false)
+
+  const { notifications } = useAdminNotifications({
+    resource_id: event.edit?.id,
+  })
+
+  const notification = notifications?.find(
+    (n) => n.event_name === "order-edit.requested"
+  )
```
srindom, line 23: praise: brilliant!

src/components/molecules/timeline-events/order-edit/confirmed-difference-due.tsx
```diff
@@ -1,50 +1,70 @@
...
+        <Button
+          onClick={() => setShowRefundModal(true)}
+          variant="ghost"
+          size="small"
+          className="w-full border border-grey-20 mb-xsmall text-rose-50"
+        >
+          Refund
           {formatAmountWithSymbol({
-            amount: orderEdit.difference_due,
+            amount: Math.abs(order?.total - order?.paid_total),
```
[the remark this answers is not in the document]
srindom: realizing that we need to take refunded total into account. So should be `order.paid_total - order.refunded_total - order.total`

srindom (commented):
Really good work! Have a few small nits and then we are good to go!

src/components/molecules/timeline-events/order-edit/confirmed-difference-due.tsx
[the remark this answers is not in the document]
fPolic: changed [here](https://github.com/medusajs/admin/pull/737/commits/1340a23b44e303223e182ad3ab9f5cfaf8445ec6)

src/components/molecules/timeline-events/order-edit/confirmed-difference-due.tsx
fPolic: Added a generic timeline event for refunds that is rendered always on top of the list if there is refundable amount in the last commit, let me know what you think :)

srindom (approved):
LGTM!

PR #738, opened 2022-10-18
fix: add loading animation to modals

kasperkristensen:
**What**
- Adds missing loading/submitting animation to "Save" buttons of several modals that had async submissions.

Closes CORE-440

### DIFF
--- a/src/domain/products/components/sales-channels-modal/index.tsx
+++ b/src/domain/products/components/sales-channels-modal/index.tsx
@@ -6,6 +6,7 @@ import LayeredModal, {
   LayeredModalContext,
 } from "../../../../components/molecules/modal/layered-modal"
 import AvailableScreen from "./available-screen"
+import { SalesChannelsModalContext } from "./use-sales-channels-modal"
 
 type Props = {
   open: boolean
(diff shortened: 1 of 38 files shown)


src/domain/products/components/sales-channels-modal/index.tsx
```diff
@@ -6,7 +6,6 @@ import LayeredModal, {
   LayeredModalContext,
 } from "../../../../components/molecules/modal/layered-modal"
 import AvailableScreen from "./available-screen"
-import { SalesChannelsModalContext } from "./use-sales-channels-modal"
```
olivermrbl, line 9: todo: This seems to be breaking the site and should not be removed

olivermrbl (changes requested):
LGTM - great work, just a single todo!

src/domain/products/components/sales-channels-modal/index.tsx
kasperkristensen: Ooops, I don't know how that happened. It should be good now!

olivermrbl (approved):

PR #744, opened 2022-10-20
feat: Add analytics to admin

kasperkristensen:
**Note**
> Requires [#2442](https://github.com/medusajs/medusa/pull/2442)

**What**
- Adds the ability to gather user insights from the admin dashboard. 
- Adds an error boundary to the application, to handle page-breaking errors more gracefully, and also collect data on these errors.

**How**
- Adds an AnalyticsProvider, which uses `Analytics.js` from Segment. 
- The provider exposes methods to track the number of products, number of orders, number of discounts, used regions and used currencies.
- On initialization, the provider will also track The store name and number of users. 

**Note**
- No data is collected in a development environment.
- The user can always opt out of sharing usage insights with us. Both when they are initially asked for permission, and if they change their mind later, it can also be turned off in `Personal Settings` under `Usage insights`.
- You can also turn off analytics for the entire store by including not including the feature flag in your server's environment.

Closes CORE-659, CORE-658, CORE-657


src/components/organisms/error-boundary/index.tsx
```diff
@@ -0,0 +1,164 @@
...
+const errorMessage = (status?: number) => {
+  if (!status) {
+    return "An unknown error occured"
+  }
+
+  return {
+    401: "You are not logged in",
+    403: "You do not have permission perform this action",
+    404: "Page was not found",
+    500: "An unknown server error occured",
```
pKorsholm, line 138: nit(non-blocking): should we have a default value for these rather than them being undefined if they fail?

pKorsholm (commented):
LGTM! very nice! 

I only have one nit 😄

olivermrbl (approved):
LGTM! Great work @kasperkristensen 💪

PR #745, opened 2022-10-20
feat: "payment required" timeline event

fPolic:
**What**
- refactor payment & refund timeline events (to be unrelated to order edits)

**How**
- implement "Customer payment required" timeline event the same way as the refund is displayed i.e. always shown on top of the list if the condition is met
- conditions for showing refund/ "payment required" need to include `difference_due` of an order edit if the OE exists that currently has status "requested"
  - shown refund if: `order.total + order.refunded_total - order.paid_total + requestedEditDifferenceDue <= 0`
  - shown "payment required" if:  ` order.paid_total - order.refunded_total - order.total - requestedEditDifferenceDue <= 0`

--- 

**Discussion**
This PR changes important order timeline parts so I want to discuss the motivation why is this implemented this way and if you guys would prefer an other approach: If we want to show "payment required" after confirmation, we cannot rely on `difference_due` since it will be 0 for the last confirmed edit. Also, we cannot show "payment required" per order edit since we don't have the computed difference between two order edits but rather the difference between an order edit and the latest order state.

If we consider the following example: 

1. the initial order is 50$
2. then first edit was created which added an item worth 10$ to the order and is force confirmed
- the customer now needs to pay 10$
3. another edit is created which adds an item worth 20$ to the order and a confirmation request is sent

At this moment order.total + order.refunded_total - order.paid_total === 10  and last_edit.difference_due === 20 
so in this case we show 30$ as the amount that is owed? Do you guys think that is ok?

---

FIXES CORE-724
FIXES CORE-725


srindom (commented):
LGTM!

srindom (approved):

srindom (approved):

PR #756, opened 2022-11-02
fix: CustomerGroup update operation

patrick-medusajs:
**What**
Fix CustomerGroup update operation. Match submitted payload to API request body shape.

**Test**
* Go to an existing customer group's page.
* Click on three dots at the right of its name and choose edit.
* Make any changes.
* Click on the Edit Group button.
* Expect Successful operation.

Fixes CORE-682

### DIFF
--- a/src/domain/customers/groups/customer-group-modal.tsx
+++ b/src/domain/customers/groups/customer-group-modal.tsx
@@ -1,4 +1,8 @@
-import { CustomerGroup } from "@medusajs/medusa"
+import {
+  AdminPostCustomerGroupsGroupReq,
+  AdminPostCustomerGroupsReq,
+  CustomerGroup,
+} from "@medusajs/medusa"
 import React, { useState } from "react"
 import { useForm } from "react-hook-form"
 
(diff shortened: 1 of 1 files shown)


src/domain/customers/groups/customer-group-modal.tsx
```diff
@@ -46,9 +52,19 @@ function CustomerGroupModal(props: CustomerGroupModalProps) {
...
+        metadata: meta,
+      }
+      handleSubmit(toSubmit)
+    } else {
+      const toSubmit: AdminPostCustomerGroupsReq = {
+        name: data.name,
+        metadata: meta,
+      }
+      handleSubmit(toSubmit)
+    }
```
srindom, line 67: suggestion: don't think the explicit type is necessary - so this would be a bit cleaner?
```suggestion
    const toSubmit = {
      name: data.name,
      metadata: meta,
    }
    handleSubmit(toSubmit)
```

patrick-medusajs: I debated between explicit and DRY. I didn't have a strong opinion for either, so let's go for DRY. A least, typing is now enforced on the `handleSubmit` param. I'll apply the suggestion. Thx.

olivermrbl (approved):
LGTM, nice one! 💪

olivermrbl:
Feel free to merge!

PR #757, opened 2022-11-03
Fix: Remove double invocation of persistedPrice in draft orders

pKorsholm:
**What**
- when setting a custom shipping price in the admin `persistedPrice` was invoked twice, once when setting the value in the `CurrencyInput.Amount` component and once before creating the draft order 

**Why**
- This caused the currency multiplier to be applied twice resulting in a shipping price inputted as 5.00 eur being 500.00 eur when the draft order was created. 

**Testing**
- Creating a draft order with a custom shipping price set
- Validate the network tab that the correct amount is shipped 
  - _The correct amount was shown in the summary, it was when the `submit` button was pressed the multiplier was applied a second time, thus we didn't register it since the price wasn't applied either_


olivermrbl (approved):
LGTM

PR #758, opened 2022-11-03
Fix: use day of month rather than day of week

pKorsholm:
**What**
- Show the correct date in order detail

**How**
- update the formatting string to display day of month, previously it displayed day of week by number

**Testing**
- open an order with a date placed > 7 and verify that it has the correct date

https://momentjscom.readthedocs.io/en/latest/moment/04-displaying/01-format/

Fixes CORE-694

### DIFF
--- a/.changeset/new-maps-breathe.md
+++ b/.changeset/new-maps-breathe.md
@@ -1,5 +1,0 @@
----
-"@fake-scope/fake-pkg": patch
----
-
-Fix: use day of month rather than day of week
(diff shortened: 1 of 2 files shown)


.changeset/new-maps-breathe.md
```diff
@@ -0,0 +1,5 @@
+---
```
srindom, line 1: nit: delete as we don't use changesets in admin

srindom (approved):
LGTM w. small nit; feel free to merge!

PR #759, opened 2022-11-03
fix: Maximum depth error in variant price form

kasperkristensen:
**What**
- Fixes an error where the PriceForm would throw a Max. depth error, when tax-inclusive prices are enabled.

**Minor**
- Add query invalidation to store query, when a currency has its tax inclusivity setting updated. When you would update a currency it would not be reflected in the price form unless you performed a hard refresh, as the Store query is not updated on currency updates., and `store.currencies` is used to display which currency prices can be set for products.

Resolves CORE-670


olivermrbl (approved):
LGTM!

PR #763, opened 2022-11-04
fix: draft order and order display the item price incorrectly when includes taxes

adrien2p:
The server return now the totals on each item of and order and cart. Use this value to display the correct amount in the order and draft order details and do not rely on the taxes locally.

depends on https://github.com/medusajs/medusa/pull/2546

FIXES CORE-687


srindom (approved):
LGTM!

PR #767, opened 2022-11-14
Fix: Quantities in fulfillment item table

pKorsholm:
**What**
- Subtract the already fulfilled amount from the fulfillable amount
- filter order items based on fulfilll-ability

Fixes CORE-437


kasperkristensen (approved):
LGTM

PR #802, opened 2022-11-24
chore: Remove @reach/router dependency and housekeeping

kasperkristensen:
**What**
- Remove @reach/router and all references (but keep compact for anyone that may have extended their admin)
- General housekeeping in the areas that have been touched as part of the above action


.vscode/settings.json
```diff
@@ -8,5 +8,8 @@
   "files.exclude": {
```
olivermrbl, line 8: comment: I think this file should be omitted from source control. Not relevant for this PR, though.

olivermrbl (approved):
Very nice @kasperkristensen. Great round of cleaning here as part of it 💪

.vscode/settings.json
kasperkristensen: Weird that it got in since its in the .gitignore, but I will remove it

echocrow: may be debatable, but it's actually good practice to commit this into source control, to ensure linting, formatting, and some common inclusions/exclusions are synced across contributors (at least amongst those using the same editor).

to clarify though, this settings file should only include linting/formatting related settings, and other stylistic'esque settings that should be consistent across developers. it should not include personal preferences like theming etc.

some repos, e.g. [tRPC](https://github.com/trpc/trpc/blob/main/.vscode/settings.json), also use it to give the top bar a tiny bit of styling, to help distinguish editors for those working across multiple projects.

(agreed it's not relevant to this PR though - I originally only snug it into to make the migration & refactor less tedious)

PR #803, opened 2022-11-24
fix: Error on product duplication

kasperkristensen:
**What**
- Fixes an issue where duplicating products with region prices would throw an XOR error.
- Fixes an issue where fields of the duplicated product and its variants were not copied over to the new product.

Fixes CORE-831


src/components/templates/product-table/use-copy-product.ts
```diff
@@ -10,51 +13,125 @@ const useCopyProduct = () => {
...
+      base.type = {
+        id: product.type.id,
+        value: product.type.value,
+      }
+    }
+
+    base.status = ProductStatus.DRAFT
+
+    // Allows the user to duplicate the same product multiple times without having to rename and change the handle of the copies
+    const uuid = v4().slice(-5)
```
olivermrbl, line 129: praise: Neat!

olivermrbl: suggestion(non-blocking): Use timestamp to make the suffix more sensible

olivermrbl (approved):
LGTM! At some point, we should add a copy endpoint to avoid doing these complex computations client-side.

PR #804, opened 2022-11-24
feat: publishable api keys UI

fPolic:
**What**
- publishable API keys management UI implementation

**TODO**
- [ ] SC management (in a separate PR, depends on a core PR)


src/components/molecules/modal/SideModal.tsx
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * Side modal displayed as right drawer on open.
+ */
+function SideModal(props: SideModalProps) {
+  const { isVisible, children, close } = props
+  return (
+    <AnimatePresence>
+      {isVisible && (
+        <>
+          <motion.div
```
kasperkristensen, line 20: question: Big fan of Framer Motion, but I am not sure if its usage here outweighs the bundle size it adds. This could potentially be replaced with CSS animations and a hook to delay unmounting. But if we are going to be doing more animations/improve the smoothness of those we already have I think it could make sense to start using it. 

One suggestion might be to use `m` that is exported from `framer-motion` instead of the full-blown `motion`, or lazy load the animations, as using`motion` will slow the initial page load a fair bit on bad connections.

kasperkristensen (commented):
LGTM! Just a couple of small nits, mostly related to not needing to import React anymore. 

Holding off with the "Approve", as I think we should hold off on merging this to develop, at least until all core PRs for PK has been merged to develop, as it is a bit tricky to run this as is for now.

src/components/molecules/modal/SideModal.tsx
fPolic: I agree it could be implemented with a basic transition but my motivation for adding the library was that we finally start to use it in the project.

The library does increase bundle size (which I don't think it's a critical issue for the admin as a non customer-facing app) but in return we get better transitions and a nice API for writing them. I would suggest that we keep the library and implement the optimisations you suggested if needed.

srindom:
@kasperkristensen - will you give the updates another round of review? :)

fPolic:
I think this should be OK to merge (if no additional changes are requested) since the core PR is merged and we have a feature flag guard

srindom (approved):
LGTM

PR #810, opened 2022-11-26
fix: Invalid hook use and cleans up dependency array

kasperkristensen:
**What**
- Fixes invalid hook use inside of a `useMemo`.
- Cleans up dependency array of memorized events.


olivermrbl (approved):
LGTM!

PR #812, opened 2022-11-27
fix: Remove ReactJson from draft order details

olivermrbl:


src/domain/orders/draft-orders/details.tsx
```diff
@@ -376,10 +376,9 @@ const DraftOrderDetails = () => {
                         </span>
                       </span>
                       <div className="flex flex-grow items-center mt-4">
-                        <ReactJson
-                          name={false}
-                          collapsed={true}
-                          src={method?.data}
+                        <JsonViewer
+                          rootName={"shipping_method"}
+                          value={method?.data}
```
kasperkristensen, line 381: Since it was collapsed before, do we not want to keep that? If so we could add this:

```suggestion
                        <JsonViewer
                          rootName={"shipping_method"}
                          value={method?.data}
                          defaultInspectDepth={0}
```

kasperkristensen (approved):
LGTM! With one comment on if we should default the view to being collapsed.

src/domain/orders/draft-orders/details.tsx
olivermrbl: Nice catch, will add :)

PR #818, opened 2022-11-28
chore: specify node versions in nvmrc & package json

jyash97:
This helps when a developer has multiple node versions installed and since different apps supports different node version specifying versions helps.


jyash97:
@olivermrbl let me know if this make sense

Issue #790, opened 2022-11-16
[Bug]: `Add Discount` when you open, it minimizes the open dropdown.


LWJerri:
Why don't we make the first dropdown open by default? It's not very convenient for admins to see it change state from open to closed before their eyes and wait for it to be clicked again.

URL: https://youtu.be/N0UC_Pj1w44

PR #821, opened 2022-12-01
FIX: Discount form first accordion item open by default

antonijoDod:
This is fix for: 
[Bug]: Add Discount when you open, it minimizes the open dropdown. #790 

I have just changed the default value in Accordion and remove unused additionalOpen prop.


kasperkristensen (approved):
LGTM thanks @antonijoDod

PR #822, opened 2022-12-02
fix: remount of OE modal

fPolic:
**What**
- the issue manifested in the order timeline which displayed (multiple) "created" order edit events

**Why**
- when we open the OE modal, a POST request is dispatched that will create a new OE on the order - this is done in an `useEffect` hook on mount
- if the newly upgraded React 18 is run in [strict mode](https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#updates-to-strict-mode), mount effects will be called twice (i.e. each component will be mounted, unmounted and mounted again)
- this will cause the creation of two order edits, and while the second one is being deleted when the modal is closed, first one still exists

**How**
- the fix is to only dispatch POST request on the first mount

---

FIXES CORE-840

### DIFF
--- a/src/domain/orders/edit/modal.tsx
+++ b/src/domain/orders/edit/modal.tsx
@@ -413,8 +413,6 @@ function OrderEditModalContainer(props: OrderEditModalContainerProps) {
     hideModal()
   }
 
-  console.log("rendered")
-
   if (!orderEdit) {
     return null
   }
--- a/src/domain/orders/edit/modal.tsx
+++ b/src/domain/orders/edit/modal.tsx
@@ -1,4 +1,4 @@
-import React, { useContext, useEffect, useState } from "react"
+import React, { useContext, useEffect, useRef, useState } from "react"
 import { Order, OrderEdit, ProductVariant } from "@medusajs/medusa"
 import {
   useAdminCreateOrderEdit,
@@ -172,10 +172,8 @@ function OrderEditModal(props: OrderEditModalProps) {
     isLoading: isRequestingConfirmation,
   } = useAdminRequestOrderEditConfirmation(orderEdit.id)
 
-  const {
-    mutateAsync: updateOrderEdit,
-    isLoading: isUpdating,
-  } = useAdminUpdateOrderEdit(orderEdit.id)
+  const { mutateAsync: updateOrderEdit, isLoading: isUpdating } =
+    useAdminUpdateOrderEdit(orderEdit.id)
 
   const { mutateAsync: deleteOrderEdit } = useAdminDeleteOrderEdit(orderEdit.id)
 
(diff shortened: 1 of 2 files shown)


src/domain/orders/edit/modal.tsx
```diff
@@ -367,43 +365,41 @@ type OrderEditModalContainerProps = {
   order: Order
 }
 
+let isRequestRunningFlag = false
```
adrien2p, line 368: Suggestion : can’t we use the isLoading from the hook instead?

src/domain/orders/edit/modal.tsx
```diff
@@ -367,43 +365,41 @@ type OrderEditModalContainerProps = {
...
-        },
-        onError: () => {
-          notification(
-            "Error",
-            "There is already active order edit on this order",
-            "error"
-          )
-          hideModal()
-        },
+    isRequestRunningFlag = true
```
adrien2p, line 386: Suggestion : instead of creating an async function that is immediately called. Why don’t we call the create and then .then .catch .finally (finally depending on the previous comment). Wdyt?

src/domain/orders/edit/modal.tsx
```diff
@@ -367,43 +365,41 @@ type OrderEditModalContainerProps = {
...
+    ;(async function () {
+      try {
+        const { order_edit } = await createOrderEdit({ order_id: order.id })
+        setActiveOrderEdit(order_edit.id)
+      } catch (e) {
+        notification(
+          "Error",
+          "There is already active order edit on this order",
```
adrien2p, line 394: Suggestion : there is already an active…

adrien2p: Question : can this happen? Since we are checking if the order edit id exists and return in that case?

adrien2p (changes requested):

src/domain/orders/edit/modal.tsx
fPolic: It won't be the same behaviour, since `isLoading` is set asynchronously, the other request will be fired.

src/domain/orders/edit/modal.tsx
fPolic: yes - `activeOrderEditId` represents OE that the user selected for editing(to be showed in the modal), while failed request means there is already an edit in the DB that is created/requested

src/domain/orders/edit/modal.tsx
adrien2p: Ok got it. Thanks for the info 💪

srindom (approved):
Brilliant LGTM :)

srindom:
@adrien2p - you need to re-review before we can merge :)

PR #824, opened 2022-12-05
fix: creating a claim fail when claim_items is not an array

adrien2p:
**What**
Prevent claim_items to be iterated when claim items are not set. Also, fix timeline which for a claim was not returning the promise from the cancellation action and therefore always return a success notification even when the request failed.

FIXES CORE-846


srindom (approved):
LGTM!

PR #835, opened 2022-12-14
fix: remove local dependency

kasperkristensen:
**What**
- The medusa-cli dependency was locked to a local version in the latest release. This PR fixes the issue, so the dependency is pulled from npm instead.


olivermrbl (approved):
LGTM!

PR #843, opened 2022-12-19
Fix popunder of Edit Variant button

### LINKED ISSUE #842: Edit Variant Button glitched
## Bug report
When trying to edit a variant, the radix button is put under the next variant, so the UI becomes very ugly.

### Describe the bug
The `z-index` set to the radix button does not work properly, because the parent element is neither `position: relative` nor does it have a own `z-index` set, so the `z-index` of the parent div is automatically higher than the highest child `z-index`, so our radix button is not shown over the elements.

### System information

Medusa version (including plugins): demo.medusajs.com, so latest.
Node.js version: demo.medusajs.com
Database: demo.medusajs.com
Operating system: Windows 10
Browser (if relevant): Firefox (latest) and Chrome (latest)

### Steps to reproduce the behavior

1. Go to [demo.medusajs.com](https://demo.medusajs.com/a/products/prod_01GMD3C4K9STFS7MAA0V8G9K6A)
2. Click on Variants -> Edit Variants
3. Click a random three-point menu
grafik
### Expected behavior

grafik

AltayAkkus:
Fixes #842

### DIFF
--- a/src/domain/products/edit/sections/variants/edit-variants-modal/variant-card.tsx
+++ b/src/domain/products/edit/sections/variants/edit-variants-modal/variant-card.tsx
@@ -118,7 +118,7 @@ export const VariantCard = ({
       ref={preview}
       data-handler-id={handlerId}
       className={clsx(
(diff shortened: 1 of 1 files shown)


AltayAkkus:
Why was this closed, it fixes #842 ?

PR #844, opened 2022-12-20
fix: OE "accepted" event

fPolic:
**What**
- fix the "Order accepted" timeline event
- fix hooks conditional rendering and unnecessary requests

---

FIXES CORE-875

PR #845, opened 2022-12-20
fix: OE "declined" timeline event

fPolic:
**What**
- fix the "Order declined" timeline event declined reason display
- properly fetch user/customer that declined the edit

---

FIXES CORE-883

### DIFF
--- a/src/components/molecules/timeline-events/order-edit/declined.tsx
+++ b/src/components/molecules/timeline-events/order-edit/declined.tsx
@@ -4,6 +4,7 @@ import React from "react"
 import { OrderEditEvent } from "../../../../hooks/use-build-timeline"
 import XCircleIcon from "../../../fundamentals/icons/x-circle-icon"
 import EventContainer from "../event-container"
+import { isDeclinedByUser } from "../../../../domain/orders/edit/utils/user"
 import { ByLine } from "."
 
 type EditDeclinedProps = {
@@ -13,7 +14,7 @@ type EditDeclinedProps = {
 const EditDeclined: React.FC<EditDeclinedProps> = ({ event }) => {
   const { order_edit: orderEdit } = useAdminOrderEdit(event.edit.id)
 
-  const declinedByAdmin = event.edit.declined_by?.startsWith("usr")
+  const declinedByAdmin = isDeclinedByUser(event.edit)
 
   const { user } = useAdminUser(event.edit.declined_by as string, {
     enabled: declinedByAdmin && !!event.edit.declined_by,
--- a/src/components/molecules/timeline-events/order-edit/declined.tsx
+++ b/src/components/molecules/timeline-events/order-edit/declined.tsx
@@ -23,9 +23,7 @@ const EditDeclined: React.FC<EditDeclinedProps> = ({ event }) => {
     enabled: !declinedByAdmin && !!event.edit.declined_by,
   })
 
-  const note = declinedByAdmin
-    ? orderEdit?.internal_note
-    : orderEdit?.declined_reason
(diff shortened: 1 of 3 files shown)


src/components/molecules/timeline-events/order-edit/declined.tsx
```diff
@@ -1,25 +1,44 @@
...
 
 type EditDeclinedProps = {
   event: OrderEditEvent
 }
 
 const EditDeclined: React.FC<EditDeclinedProps> = ({ event }) => {
-  const { customer } = useAdminCustomer(event.edit.declined_by as string)
+  const { order_edit: orderEdit } = useAdminOrderEdit(event.edit.id)
+
+  const declinedByAdmin = event.edit.declined_by?.startsWith("usr")
```
adrien2p, line 16: suggestion: I ve seen that check twice. Should we move "usr" or "cus" outside of here as const util for the order edit events to be used when needed?

src/components/molecules/timeline-events/order-edit/declined.tsx
```diff
@@ -1,25 +1,44 @@
...
+
+  const { user } = useAdminUser(event.edit.declined_by as string, {
```
adrien2p, line 18: suggestion: Since declined_by can be undefined, here the request could pass an undefined id isn't it? should we do that only if the value exists. the risk is that the server throws on invalid id instead of returning a random user ahah

src/components/molecules/timeline-events/order-edit/declined.tsx
```diff
@@ -1,25 +1,44 @@
...
+    enabled: declinedByAdmin && !!event.edit.declined_by,
+  })
+
+  const { customer } = useAdminCustomer(event.edit.declined_by as string, {
```
adrien2p, line 22: suggestion: same as previous

adrien2p (changes requested):

src/components/molecules/timeline-events/order-edit/declined.tsx
fPolic: the request won't fire if `!!event.edit.declined_by === false` (the `enabled` prop on the line below)

src/components/molecules/timeline-events/order-edit/declined.tsx
fPolic: where is the second check, [this one](https://github.com/medusajs/admin/pull/844/files#diff-c09938aa0b5e0acfc98f5670a62a66485da4ba9de87349c0afff166ec94579f9R14)?

src/components/molecules/timeline-events/order-edit/declined.tsx
adrien2p: from the jsdoc it says that the enabled only prevent from refetching no? `Set this to `false` to disable automatic refetching when the query mounts or changes query keys.` So I thought the first one would be fired but none after my bad 😂

src/components/molecules/timeline-events/order-edit/declined.tsx
adrien2p: yep, just to ensure that there is no type when someone add a new check like this one or if we change the value

kasperkristensen (approved):
LGTM!

srindom (approved):
LGTM! Think @adrien2p's idea about creating a util for determining customer/user is good but non-blocking. 

Feel free to merge :)

kasperkristensen:
@adrien2p can you remove your block? 😄

adrien2p (approved):

PR #846, opened 2022-12-25
chore: move loadingContainer to atoms

### LINKED ISSUE #575: Move LoadingContainer to atoms
**What**
Component which is used but not grouped according to the components structure of the folder

**How**
- Move components/loading-container to atoms
- Update imports

Bibek99:
**What**
Component which is used but not grouped according to the components structure of the folder

**How**
- Move components/loading-container to atoms
- Update imports

Fixes #575


Bibek99:
@olivermrbl  It would be great if you could review this pull request.

PR #852, opened 2023-01-04
fix: Product tags overflowing in product view

StephixOne:
### What & Why

Fixes [Issue#850](https://github.com/medusajs/admin/issues/850), many product tags overflowing the container. Screenshot in issue for bug.

### How

Adds a wrapping rule to the container and some vertical padding between tags.

image

Resolves CORE-947.

Reopened with correct base.


olivermrbl (approved):
LGTM, nice one @StephixOne! 💪

PR #853, opened 2023-01-04
feat: Add additional confirmation on some deletions

StephixOne:
**What**

Added the option to require extra confirmation in deletion prompts that use `useImperativeDialog`, to prompt the user to confirm deletion by additionally inputting the name of the entity they're aiming to delete.

image
image

**How**

Added two new optional arguments to `useImperativeDialog`:
- `extraConfirmation: boolean` — whether an extra confirmation should be required
- `entityName: string` — the name of the entity the deletion is for, to be input in the prompt

By default, this is opt-in functionality, which I've implemented when deleting Regions and Sales Channels, as POC. For the latter, I reworked how the deletion prompt works, to make use of `useImperativeDialog`, instead of handling it all within the component.

**Why**

Deleting some entities can be more destructive or require a lot more work to set back up, so it would make sense to triple-check the action before taking it.

Resolves CORE-328.


src/hooks/use-imperative-dialog.tsx
```diff
@@ -51,13 +74,17 @@ type ImperativeDialogProps = {
   text: string
   confirmText?: string
   cancelText?: string
+  extraConfirmation?: boolean
+  entityName?: string
 }
```
pKorsholm, line 79: nit(non-blocking): There is an edge case where if no `entityName` is passed but `extraConfirmation` is true it's possible to just confirm the modal by default: 
image

it's possible to prevent by a bit of typing like this: 
```
type ImperativeDialogProps =
  | {
      heading: string
      text: string
      confirmText?: string
      cancelText?: string
    } & (
      | {
          extraConfirmation: true
          entityName: string
        }
      | {
          extraConfirmation: undefined
          entityName: undefined
        }
    )

```

pKorsholm (commented):
Very nice! I have a minor nit but overall I love the look and feel.  😄

src/hooks/use-imperative-dialog.tsx
StephixOne: Ah neat, decent way of typing a required-together type relationship without a sub-object. Will update.

StephixOne: Done.

olivermrbl (approved):
LGTM, nice work man! 💪

riqwan (approved):
That is some clean work right there! 🎉

src/hooks/use-imperative-dialog.tsx
```diff
@@ -46,18 +69,31 @@ const DeleteDialog = ({
   )
 }
 
-type ImperativeDialogProps = {
-  heading: string
-  text: string
-  confirmText?: string
-  cancelText?: string
-}
+type ImperativeDialogProps =
```
pKorsholm, line 72: Thought: no need to change anything now, but something to keep in mind if we start adding states and possible combinations to this or to other components following this pattern, we could consider creating a type for each variant of options and combine them to improve readability. i.e. 
```
ImperativeDialogExtraConfirmationProps = 
      | {
          extraConfirmation: true
          entityName: string
        }
      | {
          extraConfirmation: undefined
          entityName: undefined
        }

ImperativeDialogProps =
  | {
      heading: string
      ...
     } & 
     ImperativeDialogExtraConfirmationProps 
```

pKorsholm (approved):
LGTM! 

I have something I think we should keep in mind, but this looks great! 😄

PR #854, opened 2023-01-05
feat: Initial sidebar redesign

StephixOne:
**What/How/Why?**

Very first stage of sidebar redesign.

- Moved the avatar/user menu from the top bar to the sidebar (and to its own component in the process)
- Changed the store name display 
- Slight redesign of the menu items, dropped the purple from active items, changed the greys a little, made the active item background borders more rounded
- Re-ordered Products/Customers/Orders, updated some of the icons
- Renamed Discounts to Promotions

image

As a sidenote, for most of my PRs updating files that contain tailwind classes: my linter re-orders the classes in what Tailwind considers the "correct" order. Makes some PRs a little more annoying to review, but have spoken to Oli and we'd like to move towards enforcing this linting rule repo-wide soon.

Closes CORE-951.

### DIFF
--- a/src/components/fundamentals/icons/cart-icon/cart-icon.stories.tsx
+++ b/src/components/fundamentals/icons/cart-icon/cart-icon.stories.tsx
@@ -1,7 +1,7 @@
 import { ComponentMeta } from "@storybook/react"
 import CartIcon from "."
 
-export default {
+const cartIconComponentMeta: ComponentMeta<typeof CartIcon> = {
   title: "Fundamentals/Icons/CartIcon",
   component: CartIcon,
   argTypes: {
@@ -12,7 +12,9 @@ export default {
       },
     },
   },
-} as ComponentMeta<typeof CartIcon>
+}
+
+export default cartIconComponentMeta
 
 const Template = (args) => <CartIcon {...args} />
 
--- a/src/components/fundamentals/icons/cart-icon/cart-icon.stories.tsx
+++ b/src/components/fundamentals/icons/cart-icon/cart-icon.stories.tsx
@@ -1,0 +1,23 @@
+import { ComponentMeta } from "@storybook/react"
+import CartIcon from "."
+
+export default {
+  title: "Fundamentals/Icons/CartIcon",
+  component: CartIcon,
+  argTypes: {
+    size: {
+      control: {
+        type: "select",
+        options: ["24", "20", "16"],
+      },
+    },
+  },
+} as ComponentMeta<typeof CartIcon>
+
+const Template = (args) => <CartIcon {...args} />
+
+export const Icon = Template.bind({})
+Icon.args = {
+  size: "24",
+  color: "currentColor",
+}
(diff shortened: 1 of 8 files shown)


src/components/fundamentals/icons/cart-icon/index.tsx
```diff
@@ -0,0 +1,49 @@
+import React from "react"
```
olivermrbl, line 1: todo: HTML attributes should be camelCased for consistency

src/components/organisms/sidebar/index.tsx
```diff
@@ -1,16 +1,16 @@
 import { useAdminStore } from "medusa-react"
```
olivermrbl, line 1: todo(this-file): Sorry, this one is on me, I should have spotted this in the design; "Orders" should be kept at the top, and "Promotions" should remain as "Discounts". The latter must be a relic from older times when we contemplated making the switch.

src/components/organisms/sidebar/index.tsx
```diff
@@ -48,15 +51,21 @@ const Sidebar: React.FC = () => {
...
+            icon={<CartIcon size={ICON_SIZE} />}
+            triggerHandler={triggerHandler}
+            text={"Orders"}
+          />
           <SidebarMenuItem
             pageLink={"/a/discounts"}
             icon={<SaleIcon size={ICON_SIZE} />}
             triggerHandler={triggerHandler}
-            text={"Discounts"}
+            text={"Promotions"}
```
olivermrbl, line 68: todo: Revert

olivermrbl (changes requested):
Strong work @StephixOne! A couple of required changes that you can blame on me. 

Not super strong in the Tailwind domain, so will need another review from one of the other guys :)

src/components/fundamentals/icons/cart-icon/cart-icon.stories.tsx
```diff
@@ -0,0 +1,23 @@
...
+  component: CartIcon,
+  argTypes: {
+    size: {
+      control: {
+        type: "select",
+        options: ["24", "20", "16"],
+      },
+    },
+  },
+} as ComponentMeta<typeof CartIcon>
```
riqwan, line 15: suggestion: `const cartIconComponentMeta: ComponentMeta<typeof CartIcon> = {...}; export default cartIconComponentMeta` something like this should give it stricter typing.

src/components/molecules/user-menu/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+              color="bg-fuschia-40"
+            />
+          </div>
+        </DropdownMenu.Trigger>
+        <DropdownMenu.Content
+          sideOffset={5}
+          className="border bg-grey-0 border-grey-20 rounded-rounded shadow-dropdown p-xsmall min-w-[200px] z-30"
+        >
+          <DropdownMenu.Item className="mb-1 last:mb-0">
+            <Button
```
riqwan, line 35: question: would it make sense to build something similar to `SidebarMenuItem` here and reuse in other dropdown menus?

riqwan (commented):
Helping me dust off my FE skills. Looks great! Added a couple of comments, only small ones. :)

src/components/molecules/user-menu/index.tsx
StephixOne: Good shout, but I'd rather check with Kasper when he's back about philosophy around these things. Have a few similar questions, just don't wanna go around refactoring things opposite to what the desired pattern might be.

src/components/fundamentals/icons/cart-icon/cart-icon.stories.tsx
StephixOne: 👍

src/components/fundamentals/icons/cart-icon/index.tsx
StephixOne: Good catch, didn't notice the discrepancy between the export and the other icon files I was comparing to! Done.

src/components/organisms/sidebar/index.tsx
StephixOne: Reverted.

riqwan (approved):
LGTM 🎉

PR #855, opened 2023-01-10
Fix delete reason button not opening dialog

josetr:


kasperkristensen (approved):
LGTM thanks for the fix!

PR #856, opened 2023-01-10
Fix "Search typing to search" typo

josetr:


kasperkristensen (approved):
LGTM thank you for the fix

PR #859, opened 2023-01-12
feat: pass order filters to export orders batch job

fPolic:
**What**
- pass filters selected in the order table header as arguments to export orders batch job

RESOLVES CORE-581


srindom (approved):
LGTM!

Have you verified that this is correctly dealt with by the server?

fPolic:
> Have you verified that this is correctly dealt with by the server?

Yes, I tried with various combinations of dates, regions, statuses etc.

kasperkristensen (approved):
LGTM!

PR #865, opened 2023-01-19
fix: position of date and month selects

kasperkristensen:
**What**
- Fixes positioning of Selects for choosing month and date using the calender component.

Fixes CORE-1016


pKorsholm (approved):
LGTM!

pKorsholm:
Maybe a very minor nit that we can handle in a different pr (since we should talk to Ludvig), it looks like you can't select past dates in the date picker when it's possible to select the dates:

Screenshot 2023-01-19 at 12 02 14

wdyt?

olivermrbl (approved):

PR #874, opened 2023-01-30
hotfix: loading price list prices in a non-edit context

kasperkristensen:
**What**
- Use `isInitialLoading` instead of `isLoading` if `isEdit` is `false`

Resolves CORE-1054


olivermrbl (approved):
LGTM, are there any other places we should be wary of? 😊

kasperkristensen:
@olivermrbl there most likely is, as we used to do this a lot, although I think most of it has been revamped in the past couple of months, we will need to test everything to make sure.

PR #876, opened 2023-01-31
fix: several issues related to order transfer and null addresses

kasperkristensen:
**What**
- Fixes an issue where the order transfer modal would throw an error if the current customer has no name and the order has no billing address. 
- Fixes an issue where edit shipping and billing address options were missing from the menu if the order has no addresses.
- Fixes an issue where the customer name would not update in the orders table after a successful ownership transfer.

Fixes CORE-1048


olivermrbl (approved):
LGTM, tested locally - nice one💪

shahednasser (commented):
@kasperkristensen Just tested it and it all works well on my end! Just one thing though: when editing a billing address that is `null`, I fill out the form normally and hit save, get a notification that the address is saved, but the address is actually still not set. Maybe because there's no address ID being passed? For reference, here's the payload:

```
{
    "first_name": "asafas",
    "last_name": "safas",
    "phone": null,
    "company": null,
    "address_1": "asfsa",
    "address_2": null,
    "city": "safas",
    "province": null,
    "country_code": "se",
    "postal_code": "3243"
}
```

olivermrbl:
@shahednasser Nice catch. I can reproduce with an integration test, so I'll create an issue for it :)

olivermrbl:
Solved in [#3153](https://github.com/medusajs/medusa/pull/3153)

PR #877, opened 2023-01-31
feat: Stock locations UI

StephixOne:
This PR implements the Inventory domain, the Locations screen within it and everything needed to create or edit stock locations and link/unlink sales channels from them. 

As part of this PR, I've also extended `useFeatureFlag` to create implicit feature flags for enabled modules (as this relies on the inventory and stock locations modules to be enabled). This makes it super easy to essentially feature-flag sections without needing actual feature flags in core, which would just match loaded modules.

**Very important:** this needs packages from [medusa#3110](https://github.com/medusajs/medusa/pull/3110) to work correctly, as that PR implements needed hooks and some types.

Resolves CORE-990 CORE-992


package.json
```diff
@@ -25,6 +25,7 @@
     "@radix-ui/react-tooltip": "^1.0.2",
     "@segment/analytics-next": "^1.43.0",
     "@tailwindcss/line-clamp": "^0.3.1",
+    "@tanstack/react-query": "^4.22.0",
     "@tanstack/react-query": "^4.22.4",
```
kasperkristensen, line 29: nit: Change this back as it works with v 4.22.4

src/components/fundamentals/icons/buildings-icon/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+const BuildingsIcon: React.FC<IconProps> = ({
+  size = "20",
+  color = "currentColor",
+  ...attributes
+}) => {
+  return (
+    <svg
+      width={size}
+      height={size}
+      viewBox={`0 0 ${size} ${size}`}
```
kasperkristensen, line 13: question: won't this result in the icon clipping on some sizes? Usually, we just leave the `viewBox` as is.

src/components/templates/address-form.tsx
```diff
@@ -22,149 +22,156 @@ export type AddressPayload = {
...
+            return (
+              <NextSelect
+                label="Country"
+                required={required}
+                value={value}
+                options={countryOptions}
+                onChange={onChange}
+                name={path("country_code")}
+                errors={errors}
+              />
```
kasperkristensen, line 171: nit: Since this field can be optional it might make sense to add an `isClearable` to it so people can reset the field.

src/domain/inventory/locations/components/address-form/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+              return (
+                <NextSelect
+                  label="Country"
+                  required={addressFieldsRequired}
+                  value={fieldValue}
+                  options={countries}
+                  onChange={onChange}
+                  name={path("country_code")}
+                  errors={errors}
+                />
```
kasperkristensen, line 139: nit: make clearable, at least if required is false

src/domain/inventory/locations/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+            <div>
+              {stock_locations?.map((stockLocation) => (
+                <LocationCard location={stockLocation} refetch={() => {}} />
+              ))}
+            </div>
+          )}
+        </div>
+      </div>
+      <Fade isVisible={createLocationState} isFullScreen={true}>
+        <NewLocation onClose={closeLocationCreate} refetch={() => {}} />
```
kasperkristensen, line 53: if refetch is not used should we just remove the prop?

kasperkristensen (approved):
LGTM with some very minor nits, that would be great to get in before we merge 🚀

src/domain/inventory/locations/index.tsx
StephixOne: Ah yes, good catch, I shoulda left a TODO comment here when I left this intermediate state whilst figuring out refetching.

src/components/fundamentals/icons/buildings-icon/index.tsx
StephixOne: Probably. I don't know why I changed the `viewBox` here and not on other icons I added 😅

src/domain/inventory/locations/new/index.tsx
```diff
@@ -0,0 +1,192 @@
...
+    },
+    reValidateMode: "onBlur",
+    mode: "onBlur",
+  })
+  const { handleSubmit, formState } = form
+
+  const notification = useNotification()
+  const { isFeatureEnabled } = useFeatureFlag()
+
+  const { mutate } = useAdminCreateStockLocation()
```
fPolic, line 57: question/suggestion: can we use `mutateAsync` here as well, so we use `async/await` throughout `handleSubmit` which would be a bit cleaner and more readable

src/domain/inventory/locations/new/index.tsx
```diff
@@ -0,0 +1,192 @@
...
+            })
+            .catch(() => {
+              notification(
+                "Error",
+                "Location was created successfully, but there was an error associating sales channels",
+                "error"
+              )
+            })
+            .finally(() => {
+              refetch()
```
fPolic, line 91: question: what are we explicitly refetching, could this be handled with cache invalidation and autorefetch?

src/domain/inventory/locations/new/index.tsx
```diff
@@ -0,0 +1,192 @@
...
+              createSalesChannelAssociationPromise(
+                salesChannel.id,
+                stock_location.id
+              )
+            )
+          )
+            .then(() => {
+              notification("Success", "Location added successfully", "success")
+            })
+            .catch(() => {
```
fPolic, line 83: question: what if only a subset of `associateSalesChannel` calls were successful, should we revert here

fPolic (commented):
Great work @StephixOne! I've added a couple of minor comments but overall LGTM 💪

src/domain/inventory/locations/new/index.tsx
StephixOne: Nothing needed anymore, will remove this in line with a comment from Kasper above.

src/domain/inventory/locations/new/index.tsx
StephixOne: Great question. I did consider this situation and the quick decision was to allow whichever associations went through to remain, and warn the user that not all happened correctly, so they may try again for the sales channels they're missing. Would be more prudent to just revert?

fPolic: I would always prefer to have an atomic update but in this case, a proper error message could suffice

src/domain/inventory/locations/components/edit-sales-channels/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+        })
+      ),
+      ...channelsToAdd.map((channelToAdd) =>
+        addLocationToSalesChannel({
+          sales_channel_id: channelToAdd.id,
+          location_id: location.id,
+        })
+      ),
+    ])
+    // for (const channelToRemove of channelsToRemove) {
```
fPolic, line 54: todo: rm if not needed

fPolic (approved):
LGTM!

pKorsholm (approved):
LGTM

PR #878, opened 2023-02-02
hotfix: Dedup default sales channel on product creation

kasperkristensen:
**What**
- Remove duplicated default sales channel entries in product sales channel array
- Use a flag to determine if users have removed the default sales channel from the product payload to prevent it from being re-added unless the user adds it themselves

Resolves CORE-1069


olivermrbl (approved):
LGTM!

PR #879, opened 2023-02-05
fix: payment required on returns

fPolic:
**What**
- fix the condition for displaying the "payment required" timeline event

---

FIXES CORE-1081


olivermrbl (approved):
LGTM!

PR #880, opened 2023-02-06
fix: Create DraftOrder flow

pKorsholm:
**What**
- Fix nested modal error with products page that would cause claim to fail
- Support for changing regions in draft orders

**How**
- Move price calculations for new variants to the draft order flow rather than the sub-modal products screen
- Update field array when updating quantities
- Update prices when changing regions in the new-order form

Fixes CORE-1082

### DIFF
--- a/src/domain/orders/details/rma-sub-modals/products.tsx
+++ b/src/domain/orders/details/rma-sub-modals/products.tsx
@@ -11,7 +11,6 @@ import { LayeredModalContext } from "../../../../components/molecules/modal/laye
 import Table from "../../../../components/molecules/table"
 import TableContainer from "../../../../components/organisms/table-container"
 import { useDebounce } from "../../../../hooks/use-debounce"
-import { useNewOrderForm } from "../../new/form"
 
 const getProductStatusVariant = (status) => {
   switch (status) {
@@ -36,10 +35,6 @@ const RMASelectProductSubModal: React.FC<RMASelectProductSubModalProps> = ({
   onSubmit,
   selectedItems,
 }) => {
-  const {
-    context: { region },
-  } = useNewOrderForm()
-
   const PAGE_SIZE = 12
   const { pop } = useContext(LayeredModalContext)
   const [query, setQuery] = useState("")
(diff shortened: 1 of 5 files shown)


olivermrbl (changes requested):
I can't seem to update the already added items. Do you know what might cause this? Adding a screen recording below.

https://user-images.githubusercontent.com/59018053/217010046-eab6a17f-2f4f-48f3-aadd-1d2a9bdf5de7.mov

src/domain/orders/details/rma-sub-modals/products.tsx
```diff
@@ -11,7 +11,6 @@ import { LayeredModalContext } from "../../../../components/molecules/modal/laye
 import Table from "../../../../components/molecules/table"
```
olivermrbl, line 11: comment: Also fixed in #881 @kasperkristensen

olivermrbl:
@kasperkristensen Would you be able to have a look at this? 🙏

src/domain/orders/new/components/items.tsx
```diff
@@ -32,19 +32,27 @@ const Items = () => {
...
 
-  const addItem = (variants: PricedVariant[]) => {
+  const addItem = async (variants: ProductVariant[]) => {
     const ids = fields.map((field) => field.variant_id)
+
     const itemsToAdd = variants.filter((v) => !ids.includes(v.id))
 
+    const variantIds = itemsToAdd.map((v) => v.id)
+
+    const { data } = (await Medusa.variants.list({
```
kasperkristensen, line 49: nit (non-blocking): I think we should look to phase out the old API service, so I am not too keen on adding the use of it to a new file. Could we use medusajs instead?

kasperkristensen (approved):
LGTM!

src/domain/orders/new/components/items.tsx
olivermrbl: @pKorsholm Do you have any comments on this? :)

pKorsholm: changed it 😄

olivermrbl (approved):
LGTM! 😍

PR #881, opened 2023-02-06
fix: failure on null claim items

kasperkristensen:
**What**
- Fixes an issue where region_id was missing in select products RMA table
- Fixes an issue where `claim.claim_items` was iterated while being null, throwing an error.

Fixes CORE-1084


src/domain/orders/details/claim/create.tsx
```diff
@@ -355,6 +353,7 @@ const ClaimMenu: React.FC<ClaimMenuProps> = ({ order, onDismiss }) => {
             />
             {returnShippingMethod && (
               <RMAShippingPrice
+                inclTax={false}
```
olivermrbl, line 356: question: If we default to tax-exclusive, should we add a note/tooltip on it?

olivermrbl (changes requested):
It seems the refund amount on the return order (created from claiming) is incorrect. And so is the shipping method amount:
CleanShot 2023-02-06 at 14 49 45

On the above order, I set a custom price of $5

src/domain/orders/details/claim/create.tsx
olivermrbl: Re flag, since it hasn't been there previously, let's keep that for a follow-up iteration.  

Re modal issues, I think that makes sense. Let's not make too many duct tape solutions now, if we have a larger revamp coming soon.

So this PR will be isolated to only fixing the null on claim items right?

PR #886, opened 2023-02-07
fix: claim menu

kasperkristensen:


olivermrbl:
Would be super helpful to display the price of return shipping when it has been selected, as well as the full refund amount:

CleanShot 2023-02-08 at 17 39 56

What do you think?

PR #888, opened 2023-02-08
fix: Use custom shipping price in DraftOrder creation flow

olivermrbl:


src/domain/orders/new/new-order.tsx
```diff
@@ -54,9 +54,10 @@ const NewOrder = ({ onDismiss }: NewOrderProps) => {
         shipping_methods: [
           {
             option_id: data.shipping_option.value,
-            price: data.custom_shipping_price
-              ? data.custom_shipping_price
-              : undefined,
+            price:
+              typeof data.custom_shipping_price === "number"
```
adrien2p, line 58: 🙏 nice catch. It is important when we work with numbers check the typeof it as 0 is considered as false

adrien2p (approved):

PR #889, opened 2023-02-08
fix: running effect after OE modal closes

fPolic:
**What**
- I've noticed that this started to occur after we updated `react-query` recently so I presume that async flow is different now causing `activeOrderEditId` to be set to `undefined` before the last effect call in the modal is triggered. Fixed by explicitly removing active edit id after the modal is hidden.

---

FIXES CORE-1047

### DIFF
--- a/src/domain/orders/edit/modal.tsx
+++ b/src/domain/orders/edit/modal.tsx
@@ -418,7 +418,7 @@ function OrderEditModalContainer(props: OrderEditModalContainerProps) {
   }, [activeOrderEditId])
 
   const onClose = () => {
-    setActiveOrderEdit(undefined)
+    // setActiveOrderEdit(undefined) -> context will unset active edit after flag toggle
     hideModal()
   }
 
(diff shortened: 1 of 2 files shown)


src/domain/orders/edit/modal.tsx
```diff
@@ -418,7 +418,7 @@ function OrderEditModalContainer(props: OrderEditModalContainerProps) {
   }, [activeOrderEditId])
 
   const onClose = () => {
-    setActiveOrderEdit(undefined)
+    // setActiveOrderEdit(undefined) -> context will unset active edit after flag toggle
```
olivermrbl, line 421: question: Do you think we should keep this in?

olivermrbl (approved):
Works! LGTM 💪

src/domain/orders/edit/modal.tsx
fPolic: Since the flow is a bit unconventional I think it's ok to leave this comment here for clarification

PR #895, opened 2023-02-13
fix(discounts): race condition causing empty regions list to be rendered sometimes

fPolic:
**What**
- logic for rendering the form with regions dropdown was relying on `setTimeout` of 100ms but in a case where a response with regions would take longer than that, an empty dropdown would be rendered

---

FIXES CORE-1099


pKorsholm (approved):
LGTM! Nice catch 💪

srindom (approved):
LGTM!

PR #897, opened 2023-02-13
fix: Claim and Receive Return menus

kasperkristensen:
**What**
- [x] Fix custom shipping prices not being part of requests
- [x] Fix custom refund amount not being part of requests
- [x] Revamp ClaimMenu
- [x] Revamp ReceiveReturnMenu  

**Testing**
- [x] Adds `@testing-library/react`
- [x] Adds `vitest`
- [x] Add tests for all form components used in revamped menus 

**Also**
- [x] Removes unused JS contexts and changes some files to `.ts` so we can wrap our test environment in the different providers we use. (Only works with `esm`)
- [x] Updates eslint config to allow `vitest` globals
- [x] Updates prettier config to use tailwindcss plugin to sort classes and pick up duplicate/conflicting classes

Depends on https://github.com/medusajs/medusa/pull/3276


src/components/molecules/timeline-events/return.tsx
```diff
@@ -38,24 +38,12 @@ const Return: React.FC<ReturnRequestedProps> = ({ event, refetch }) => {
     })
   }
 
-  const { mutateAsync: receiveReturn } = useAdminReceiveReturn(event.id)
+  const args = buildReturn(event, handleCancel, openReceiveReturnMenu)
```
olivermrbl, line 41: suggestion(non-blocking): Rename to something more descriptive
```suggestion
  const eventContainerArgs = buildReturn(event, handleCancel, openReceiveReturnMenu)
```

olivermrbl (changes requested):
I could not get far in the claims flow due to an issue with the return reasons - added a SS below. Therefore, this review is scoped to UI unrelated to RMA.

## Attempting to request a claim without a shipping address on order:
CleanShot 2023-02-21 at 20 08 01@2x

## Attemtping to select a return reason: 
CleanShot 2023-02-21 at 20 10 08@2x

## Hover on avatar dropdown has funky outline:
CleanShot 2023-02-21 at 20 25 24@2x

kasperkristensen:
@olivermrbl can you try again with an order without a shipping address?

olivermrbl:
@kasperkristensen Couple of additional things. A todo (bottom one) keeping me from doing anything after having requested a claim. 

**todo: I can no longer delete a requested claim:**
CleanShot 2023-02-22 at 12 42 37

**nit(non-blocking): Select has funky background on hover:**
CleanShot 2023-02-22 at 12 44 36

todo: Cannot load order page after having requested a claim with a custom refund amount:
CleanShot 2023-02-22 at 12 46 19

Above error is thrown when attempting to read `return_order.status` in the `use-build-timeline.tsx` file:
CleanShot 2023-02-22 at 12 47 59

olivermrbl:
todo: **Refundable amount is off in "Receive return" menu:**
CleanShot 2023-02-22 at 12 55 21

Should we just delete the "refundable" column? Does not make much sense when you have the "Refund amount" in the summary.

praise: I absolutely love this summary:
CleanShot 2023-02-22 at 12 58 16

olivermrbl:
nit(non-blocking): Raw JSON blob has weird pattern when collapsed:
CleanShot 2023-02-22 at 13 04 38

kasperkristensen:
@olivermrbl I'll pinpoint the version of Tailwind which results in the weird top margin, this does show that we have a lot of cleanup to do in our styling, as this again is a result of the newer version of Tailwind not being as lenient with invalid/conflicting styles.

Refundable is not off per se, it is listing how much is refundable for the line item (`item.refundable`), so it's not correlated with the refund that is about to occur. I am good with removing it, but I am guessing that we included it in the design for some reason initially, and not sure if we would be removing some info used by admins? But I agree that it is very confusing.

I'll fix the other things you pointed out

olivermrbl:
nit: Cancel claim should be disabled when `claim.fulfillment_status === "shipped"`
image

olivermrbl (approved):
LGTM, absolutely fantastic work @kasperkristensen!

Quite a huge PR to get through. To be honest, I did not review every file in detail, but I tested all the flows thoroughly, and they work as expected. 

Let me know, if there's anything specific you want me to take an extra look at. Otherwise, fine to merge this in as is!

🚢 🚢 🚢

pKorsholm (approved):
LGTM

PR #900, opened 2023-02-22
Fix(admin): Fix for amount not shown

pKorsholm:
**What**
- Ensure that 0-amounts are shown for requirements

**How**
- We validated using the amount as boolean value, however 0 is false. I added ternary checks instead of OR assignments to validate if value was a number, otherwise we use undefined. 

Fixes CORE-1075


olivermrbl (approved):
LGTM, nice one!

olivermrbl:
Will merge when #897 has landed in `develop`

kasperkristensen (approved):
LGTM

PR #911, opened 2023-03-03
hotfix: add skus to claim tables

kasperkristensen:
**What**

- Adds SKUs to all table views in claims flow


olivermrbl:
I am seeing an unrelated toaster when I copy to clipboard:
CleanShot 2023-03-03 at 13 59 21

kasperkristensen:
@olivermrbl I forgot to copy over the fix to the CopyButton, as it was defaulting to a submit button when used in a form. Should be good now, and is already present in the other PR. 😄

olivermrbl (approved):
Works!

